PR #26, opened 2026-05-31
feat: Add media — yt-dlp URL downloads + torrent add

IrosTheBeggar:
## Summary

Adds an in-app **"Add media"** capability that mirrors the mStream webapp: download audio from a URL via the server's yt-dlp, and add torrents (magnet or `.torrent`) with smart metadata→path detection. All of it drives **base mStream server routes** (`/api/v1/ytdl/*`, `/api/v1/torrent/*`) with `x-access-token` auth — the download/processing happens server-side and the results land in the server's library.

## What's new

**yt-dlp — "Download from URL"** (server file-explorer header):
- Paste a YouTube URL → debounced metadata preview (editable title/artist/album/year + thumbnail) → pick an output codec (from the server's `supportedAudioFiles`) → download into the directory currently in view.
- A "Downloading N…" banner (`YtdlManager`, 3s poll of `/ytdl/downloads`) that auto-refreshes the file list when a download completes in the directory you're looking at.

**Torrent tab** (Add-media sheet):
- Preflight capability gate (shows the configured client or why it's unavailable), magnet input, **`.torrent` file upload** (`file_selector` — builds on the AGP-9 toolchain where `file_picker` 11.x didn't), live destination preview → `POST /torrent/add`.

**Smart "Add Torrent" screen** (drawer):
- Client-side `parseMusicTorrentName` (6 patterns + tag-stripping) on the bencode name / magnet `dn=`, optionally upgraded by the server's auto-detect.
- Per-library **path templates** (`{{ARTIST}}/{{ALBUM}}`, sanitized) with a legacy `artist/album` fallback; sticky once you edit the path.
- Multi-server picker (supported servers only) that reloads preflight + templates on switch.
- **Seed-existing pre-check** (`POST /seed-existing`): seed-in-place when files already exist, a partial-match sheet to point at existing locations, or fall through to a fresh add. A "Force fresh download" toggle bypasses it.

## Hardening (audit follow-up)

A two-agent audit of the rebased branch surfaced two issues worth fixing before merge, now closed:
- **Path-traversal:** `splitTorrentPath` passed raw segments to `/torrent/add`. Now every segment is run through `sanitizeTorrentSegment` and split on `/`+`\`, so `..`, backslashes, control chars, and a leading `/` are neutralized client-side (no longer relying solely on the server to re-validate).
- **ytdl server-switch:** `YtdlManager` polled the live current server, so switching servers mid-download misrouted the "Downloading…" banner. It now binds to the submitting server at `start()`, only refreshes the list while that server is in view, and clears its completion-dedup set on stop.
- **Resilience (follow-up):** a transient `daemon_error` from the seed-existing pre-check now falls through to a normal add instead of dead-ending the submit; `ytdlMetadata`/`ytdlDownloads` decode defensively (a non-object body surfaces a clean error / empty list rather than throwing).
- **Add-media polish (follow-up):** the Download tab pre-flight-probes the server (`/ping`) and disables Download with a banner when it's unreachable, instead of failing only after submit; the Torrent tab's `/preflight` is now lazy (fires only when the tab is opened); empty-variable template segments no longer yield a junk `-` folder; and an empty-`relativePath` partial match surfaces a notice instead of doing nothing.
- **Smarter name parsing (follow-up):** informed by how Lidarr/beets and the scene FLAC/MP3 standards parse releases — keeps hyphenated artists intact (`Jay-Z`), and now **deterministically parses scene dirnames** (`Pink_Floyd-The_Wall-1979-FLAC-PMS` → `Pink Floyd / The Wall / 1979`, where `_`=space and `-`=field separator), canonicalizes `VA`→`Various Artists`, strips `feat.` clauses from the album, and recognizes a far wider token set (codecs, sources, disc markers, editions, status). **52 unit tests.**

## Testing

- `flutter analyze` clean (only pre-existing repo warnings).
- **31 unit tests** for the pure path/metadata logic (`torrent_meta.dart`) — up from 20, adding path-traversal safety assertions and coverage of the `extractTorrentName` bencode parser.
- Debug APK builds (the new `file_selector` plugin links on the AGP-9 transitional toolchain).
- Rebased cleanly onto current master (post the storage-location PR #25).

## Known limitations / follow-ups (out of scope here)

- **Parser:** scene dirnames now parse deterministically, but a *space- or dot-only* name with a multi-word artist and no `-` separator (`Pink Floyd The Wall 1979`) still can't reliably tell where the artist ends and the album begins — that case may need a manual tweak (all fields editable). The more robust fix (cross-referencing the `.torrent` file list à la beets/Headphones) is noted as a follow-up.
- **Server-trust:** the user URL is host-gated to YouTube client-side and torrent paths are now sanitized, but the **server must still validate** both — the client checks are convenience, not a security boundary.
- **ytdl capability gate** is reachability-based (a `/ping` probe) — it catches an unreachable/misconfigured server, but a server that's reachable yet has yt-dlp specifically disabled is still only reported after submit.


IrosTheBeggar:
Superseded by #117, which re-lands the torrent half of this branch on current master (418 commits of drift made an in-place rebase impractical — the l10n system, BrowserToolbar, and the iroh transport all postdate this branch).

What carried over: the release-name parser + its 52 tests verbatim, the smart Add-torrent screen, and the seed-existing flow. What changed: the API layer now goes through Server.apiUri (iroh-tunnel support), the UI is fully localized (44 keys × 10 locales), the velvet-server filter is dropped (the flag no longer exists), and a junk-"()"-segment template bug found on-device is fixed with tests.

The yt-dlp half (Download-from-URL, YtdlManager, the Add-media sheet) is still only on this branch — it should get the same treatment as a follow-up. The branch is left in place as the reference for that port.

PR #33, opened 2026-06-03
fix: duplicate-id queue rows highlight and dismiss the wrong row

IrosTheBeggar:
## What

The "Up Next" queue list misbehaved whenever the queue legitimately held the **same track id more than once** (a supported case):

1. **Active highlight bled across copies** — the active-row highlight + EQ/pause badge lit up *every* row sharing the playing track's id, not just the one actually playing.
2. **Swipe-to-dismiss removed the wrong row** — swiping any duplicate deleted the *first* occurrence of that id, not the row the user swiped.

## Why

Both reduce to `MediaItem.==` being **id-based**: audio_service overrides `==` to `other is MediaItem && other.id == id`. So the two id-based lookups in `queue_list.dart` collapsed distinct rows that happened to share a track id:

- active check: `item == current`
- dismiss lookup: `live.indexOf(item)` (returns the first id match)

## How

Disambiguate by **queue position / instance identity** instead of id:

- **Active row** — compare the row index against `PlaybackState.queueIndex` (the playing slot) rather than `item == current`. `queueIndex` is set from `_backend.currentIndex` and updates off the *same* just_audio `playbackEvent` that previously drove `mediaItem`, so the highlight is just as prompt but unambiguous. The now-unused `mediaItem` source drops out of the `combineLatest` (3 streams → 2).
- **Dismiss** — resolve the row with `indexWhere((m) => identical(m, item))` instead of `indexOf(item)`, so swiping a duplicate removes that exact row. Safe because the handler mutates the queue in place (`queue.value..removeAt`), keeping `MediaItem` instances stable; the existing stale-index re-resolution at dismiss time is preserved.
- **Row key** — `ObjectKey(item)` (instance identity) instead of `ValueKey('${id}#$index')`. Collision-free because every enqueue path builds a fresh `MediaItem`, and it follows a row across reorders/removals so in-flight `Slidable` state survives an index shift.

`onTap` already used `skipToQueueItem(index)` (position-based), so it was never affected.

## Reviewer notes

- Pre-existing bug, **not** introduced by the reorder PR #32 — it was surfaced during that PR's audit.
- Scope: a single file, `lib/widgets/queue_list.dart`. `_QueueSnapshot` is private and the public `QueueList` / `QueueHeader` / `downloadQueue` signatures are unchanged, so there's no cross-file impact.
- `flutter analyze lib/widgets/queue_list.dart` → no issues; `flutter test` → all 22 tests green.
- **Not** verified at runtime on a device: `QueueList` reads the `MediaManager().audioHandler` singleton via a module-level stream that needs `AudioService.init` (platform channels), so it isn't unit-testable without the channel mocking that `TEST_PLAN.md` defers. To verify manually: queue the same track twice, confirm only the playing copy is highlighted, and swipe each copy to confirm the swiped row is the one removed.

### DIFF
--- a/lib/widgets/queue_list.dart
+++ b/lib/widgets/queue_list.dart
@@ -11,28 +11,37 @@ import '../singletons/media.dart';
 import '../theme/velvet_theme.dart';
 import '../util/media_format.dart';
 
-/// Combined snapshot of the queue, the currently-playing item, and whether
-/// playback is running — so a row can render the list, highlight the active
-/// track, and show a live EQ / paused badge over its art.
+/// Combined snapshot of the queue, the index of the currently-playing slot,
+/// and whether playback is running — so a row can render the list, highlight
+/// the active track, and show a live EQ / paused badge over its art.
+///
+/// The active row is keyed off [activeIndex] (the queue position, from
+/// PlaybackState.queueIndex) rather than by matching the playing MediaItem:
+/// MediaItem.== is id-based, so a queue that legitimately holds the same track
+/// id more than once would light up every matching row at once. A position is
+/// unambiguous.
 class _QueueSnapshot {
   final List<MediaItem> queue;
-  final MediaItem? current;
+  final int? activeIndex;
   final bool playing;
-  const _QueueSnapshot(this.queue, this.current, this.playing);
+  const _QueueSnapshot(this.queue, this.activeIndex, this.playing);
 }
 
 /// Memoised as a single broadcast subject (mirrors player_panel's _mediaPos):
 /// the queue list is rebuilt by the panel's drag AnimationController, so a fresh
-/// combineLatest per build would re-subscribe three handler streams every frame.
+/// combineLatest per build would re-subscribe its handler streams every frame.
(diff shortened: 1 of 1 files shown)


IrosTheBeggar:
Superseded by #35.

Thanks for catching this — the underlying bug is real. Closing in favour of #35 because this PR has drifted out of sync with `master`:

- **Conflicting now.** It was branched before #32 (reorder) and #34 (swipe/info) merged, so it no longer applies cleanly (`mergeable: CONFLICTING`).
- **Two of its three changes already landed.** The `indexWhere((m) => identical(m, item))` dismiss fix and the `ObjectKey(item)` row key are already on `master` via #34's audit — so those hunks would conflict for no gain.
- **The remaining change would regress a rebuild throttle.** As written it reads `queueIndex`/`playing` off the full `playbackState` stream, but `master` deliberately `distinct()`s those (playbackState re-emits several times a second on position ticks) to avoid rebuilding the reorderable queue. Merging as-is would reintroduce that jank.

#35 keeps the still-needed part — the active-highlight-by-position fix — rebased on current `master`, and gives `queueIndex` its own `distinct()` so the throttle is preserved.

PR #46, opened 2026-06-06
Full/play flavors: storage compliance + self-signed TLS (full only)

IrosTheBeggar:
Google won't accept `MANAGE_EXTERNAL_STORAGE` ("All files access") for a music player. Instead of forking, this splits the app into **two Android product flavors from one codebase**, so all future features/fixes stay shared.

## The two flavors
| | `full` | `play` |
|---|---|---|
| Distribution | Sideload / direct APK | Google Play |
| applicationId | `mstream.music` | `mstream.music.play` (distinct → coexists) |
| Storage perms | WRITE/READ/**MANAGE_EXTERNAL_STORAGE** | **none** (aapt-verified) |
| Download modes | App local, App external, **Permanent, SD card** | App local, **App external** |

The `MANAGE_EXTERNAL_STORAGE` permission now lives **only** in `android/app/src/full/AndroidManifest.xml`; `main` has none, so the `play` merged manifest is permission-clean by construction (the manifest is the authoritative compliance control).

## Compliant storage
Added an **`appExternal`** mode → app-scoped external (`Android/data/<pkg>`): more room / SD-capable, **no permission**, but removed on uninstall (same caveat as App local). The picker offers App local + App external always; Permanent/SD card only when `!isPlayBuild` (`--dart-define=VARIANT=play`).

## ⚠️ Workflow change
Because product flavors now exist, **every build/run must pass `--flavor`**:
- `flutter run --flavor full` (or `play`)
- `flutter build apk --flavor full`
- `flutter build appbundle --flavor play --dart-define=VARIANT=play`

CI (`release.yml`) is updated: on a tag it builds the **full APK** (attached to the GitHub Release for sideload) **and** the **play AAB** (for the Play Console).

## Verification
- `aapt dump permissions`: `full` has the 3 storage perms; `play` has none. applicationIds `mstream.music` vs `mstream.music.play`.
- `flutter analyze` clean · `flutter test` 34/34 · both debug flavors build under AGP 9.
- ⚠️ On-device launch check pending — the wireless test device dropped mid-run. Low risk (Dart changes are analyze-clean; build-time split verified via aapt), but worth a quick run of each flavor before merge.

## Notes / decisions captured
- Distinct `.play` applicationId (your call) — Play and sideload coexist, independent signing, no uninstall dance.
- Aside: the cast plugins pull in `ACCESS_FINE_LOCATION` — also Play-sensitive, pre-existing, not addressed here.


IrosTheBeggar:
**Update — added self-signed TLS support (full flavor only)**, building on the flavor split in this PR.

Per-server **"Allow self-signed certificate"** toggle (off by default; hidden on the Play build). The catch is that API and streaming use different TLS stacks:
- **API** (Dart `http` → `dart:io`): `SelfSignedHttpOverrides` (HttpOverrides.global, full flavor) bypasses validation **only** for hosts of servers that opted in.
- **Streaming** (just_audio → ExoPlayer → native TLS, unreachable from Dart): a native trust-all bridge via the `mstream/insecure_tls` channel, enabled when any server opts in. It lives **only** in `android/app/src/full/kotlin` — the `play` source set is a no-op, so the trust-all `X509TrustManager` never enters the Play binary.

Verified via dex grep: the `mstream/insecure_tls` channel is present in the **full** APK and **absent from play** (0 occurrences). analyze clean · 34/34 tests · both flavors build.

One nuance: the native bridge is process-global once enabled, so *streaming* trust isn't strictly per-server (the API path is). Acceptable for a power-user sideload switch. On-device functional check needs an actual self-signed server (deferred).

PR #73, opened 2026-06-17
Visualizer performance pass (ShaderEngine + shared infra)

IrosTheBeggar:
Performance audit of the visualizer stack (both engines + shared infrastructure), implemented as six focused commits. Each addresses a distinct hot-path cost; the ProjectM path is a thin passthrough to `libprojectM-4.so`, so the wins are concentrated in the custom **ShaderEngine** and the shared audio/present/pacing code.

## Commits (one per finding)

1. **Render constant-output buffer passes at a declared size.** The audio-analysis buffer passes (cyber-fuji `bufferA`, hex-marching `bufferC`) compute one `vec4` of state and write it identically to every pixel; the image pass then samples a single texel. They were running full-screen passes — at ~2400×1080 that's ~2.6M invocations each doing up to 24 `texture()` fetches to produce 4 floats, plus two full-res RGBA8 ping-pong textures (~20 MB). Added a `// === size <pass> = WxH` marker so those buffers render at `1x1`. Biggest GPU + memory win.
2. **Blit instead of a textured present pass when not crossfading.** The present pass ran a full-screen textured draw every frame even though the crossfade is only live for 0.6 s after a swap. When idle, copy `fboCurrent_` → window with `glBlitFramebuffer` and skip the program/sampler/draw setup.
3. **`eglSwapInterval(0)`.** The render thread paces to ~60 fps with `Thread.sleep` and the transcode path paces to the audio clock, but swap blocked on vsync (default interval 1) — double-throttling, and beating against 90/120 Hz panels. Make the manual pacing the single clock.
4. **Drop the per-frame defensive `eglMakeCurrent`** on the on-screen render path. The context never leaves the dedicated render thread (the encoder path already relies on this).
5. **Hoist constant uniforms out of `renderPass`.** `iChannelResolution`, `iSampleRate`, `iMouse`, `iDate`, the `iChannel` sampler-unit bindings, and `iParams` were re-uploaded every pass every frame though they never change frame-to-frame. Push them once at adopt (`primeConstantUniforms`); re-push `iParams` only when `setTuning` dirties it.
6. **Cut per-sample trig in the synthesized PCM generator.** `_generate()` runs on the UI isolate and did 7 `sin()`/sample. The sweep LFOs (<1 Hz) are hoisted to once-per-tick and the carriers use phase accumulators → 4 `sin()`/sample (~43% fewer).

## Verification
- All native translation units compile clean with the project's pinned NDK (28.2, arm64, `-Wall`).
- `flutter analyze` clean on the changed Dart.
- **Not** yet built into an APK or run on-device — compile/analyze only.

## Behavioral note
Commit 6 slightly changes the *character* of the synthesized idle signal (cleaner swept tones vs the old buzzy texture). It only affects the default synthesized-audio mode (real-audio capture is unaffected) and is purely cosmetic.

## Deliberately out of scope
- **Render-scale dial** (render offscreen at a fraction of window res, upscale in the present pass) — a real lever but a quality-changing feature, not a fix. Easy follow-up.
- ProjectM mesh size (48×36 is a sensible default).

### DIFF
--- a/android/app/src/main/cpp/shader_engine.cpp
+++ b/android/app/src/main/cpp/shader_engine.cpp
@@ -789,7 +789,7 @@ void ShaderEngine::renderFrame() {
         }
     }
 
-    // === Present pass: fboCurrent_ (with optional crossfade) → window ===
+    // === Present: fboCurrent_ → window, with optional crossfade ===
     float mixT = 1.0f;
     if (transitioning_) {
         const float t = std::chrono::duration<float>(now - transitionStart_).count()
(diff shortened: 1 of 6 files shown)


IrosTheBeggar:
Closing in favor of #74, which is the curated, on-device-tested subset of this PR.

Each commit here was A/B-tested on a Galaxy S25 (profile builds; true per-frame GPU time via `glFinish`, GL memory via `dumpsys meminfo`) and correctness-reviewed:

- **Keep (→ #74):** 1×1 buffer passes, blit present, drop per-frame `eglMakeCurrent`, hoist constant uniforms, cut per-sample PCM trig. All reviewed correct.
- **Dropped:** `eglSwapInterval(0)` — proven a no-op on a Flutter SurfaceTexture surface (#39 + re-confirmed here).

Key measured result: **no framerate change on the S25** (the visualizer isn't GPU-bound — the 4-pass hex shader renders in ~4 ms), but the 1×1-buffer change does cut **~20 MB of GL memory** (144,856 → 124,492 KB), matching its claim. #74 also adds two robustness fixes (guard the untrusted size-marker parse; refresh a stale doc comment).

🤖 via Claude Code

PR #89, opened 2026-06-28
feat(audio): attach the EQ pipeline only when the equalizer is enabled

IrosTheBeggar:
## Why
Continuation of the Play "stops playing" hardening (#85 / #86 / #87). An `AndroidEqualizer` was attached to the just_audio `AudioPipeline` on **every** Android build, even when EQ was disabled. just_audio re-activates a priority-0 audio effect on every audio-route change (Bluetooth/wired switch); on Samsung — where SoundAlive/Dolby own the effect chain — that can drop the effect slot and cut playback. Users who never touch EQ (the majority) were carrying that risk for nothing.

## What
just_audio fixes the pipeline at player construction (the effect list is construction-time only; an effect instance is bound to one player), so the player is now **plain by default** and **rebuilt with the EQ pipeline** when the user turns EQ on (and back to plain when off).

- `LocalPlaybackBackend`: mutable player + equalizer; `_buildPlayer` / `rebuildPlayer`; `supportsEqualizer` is now the stable `Platform.isAndroid` capability.
- `AudioPlayerHandler.setEqEnabled`: rebuilds the local player, carries playback (pos/index/play + shuffle/repeat) across, re-emits the backend so derived streams re-bind, re-applies saved gains (bounded), re-points the visualizer. Serialized on `_switchChain`, gated on `_initialized`.
- `restoreQueue` now runs on `_switchChain` too — closes a cold-start restore race the toggle exposed ("Loading interrupted").
- `eq_screen`: the enable toggle is always visible on Android (it's how EQ is turned on while the equalizer object doesn't exist yet); new "EQ off" state distinct from "Android only".

## Review
3-dimension adversarial review (concurrency / EQ-screen / regression). It surfaced a `_switchChain` deadlock (unbounded `eq.parameters` await on an idle queue), the restore race, and a persist-order drift — all fixed in this PR.

## Trade-off
Toggling EQ rebuilds the player → a brief re-buffer of the current track at the same spot. Inherent to just_audio's construction-time pipeline; acceptable since toggling EQ is rare and deliberate.

## Testing
- [x] `flutter analyze` clean; l10n regenerated with `eqDisabledHint`.
- [ ] On-device (Samsung) smoke test — see checklist in the PR discussion.


IrosTheBeggar:
**Audited, kept, and rebased onto post-v0.31.0 master** (single squashed commit replacing the June history — the old base predated #94–#103 and conflicted heavily).

**Audit verdict: worth merging, and worth keeping the EQ.** The motivation is squarely part of the reliability effort: an always-attached `AndroidEqualizer` re-activates as a priority-0 effect on every audio-route change, and on Samsung the SoundAlive/Dolby effect-chain fight can drop the slot and cut playback — the majority who never open the EQ were carrying that risk for nothing. This PR removes the risk for them while keeping the feature for EQ users; removing the EQ outright would trade a shipped feature for less gain than this achieves.

**Rebase adaptations** (the handler changed a lot under this PR): the rebuild's playback carry now uses `_reviveSpot()` + `_queueWithFreshUrls()`; the original restoreQueue hunk was dropped (master's #99/#103 version supersedes it).

**Adversarial re-review of the rebased result (2 lenses, 13 verify agents) confirmed 7 defects — all fixed:**
1. **(high)** A failed re-seed (offline/dead-server EQ toggle) skipped the backend re-emit *after* the old player was disposed → every handler stream deaf forever (no error recovery, no completed→stop, frozen notification). The re-emit now happens immediately after the rebuild, before any fallible await.
2. A failed re-seed lost the playback spot (empty player reads 0/0:00; snapshot persisted the loss) → now parks `_restoreSpot`, mirroring restoreQueue.
3. The old player's dispose-time events, read through the new empty player, flashed a bogus `error` to the media session mid-toggle → broadcasts gated while `_rebuilding`.
4. The WifiLock predates-patch marker's `loading` clear was unobserved (event fired pre-rebind) → fixed by the early re-emit.
5. The final seek enforced the *captured* playing state over a pause/play that landed during the reload → now honors live `_playIntent`.
6. Reset-all flipped `eqEnabled` under a still-pipelined player (audio stayed EQ'd while the screen said Off) → routed through the handler.
7. (accepted, documented) a ms-scale window where a download completing mid-rebuild can briefly desync the WifiLock — self-corrects on the next track.

`flutter analyze` clean (baseline), 126/126 tests, APK builds, deployed to the test device. **Remaining check:** toggle EQ on/off mid-playback on-device (playback should continue at the same spot with a brief re-buffer, gains applied) — the phone went to bed before I could drive it; toggling the EQ while music plays *is* the test.

IrosTheBeggar:
**On-device smoke test (S25, UI-driven) — PASS. The remaining checklist item is done:**

1. **EQ OFF baseline**: plain player, EQ screen shows the enable toggle + "Off — bypass mode" + the new eqDisabledHint, no sliders (no pipeline attached). Playback running.
2. **Toggle ON mid-playback**: player rebuilt with the pipeline — playback continued without interruption (position advanced continuously through the rebuild; the re-buffer on a downloaded track is imperceptible), the 5-band slider board rendered, and the **saved gains re-applied automatically** (+3dB @ 60Hz/14kHz — the gains-on-first-ready deferral working end-to-end).
3. **Toggle OFF mid-playback**: rebuilt back to plain — playback again continuous, sliders gone, hint back, `eqEnabled:false` persisted.
4. Zero error lines in the log across both rebuilds; no bogus media-session error flash (the `_rebuilding` broadcast gate).

Ready to merge.

PR #98, opened 2026-07-08
feat(playback): hold a WifiLock while audio streams over the network

IrosTheBeggar:
## What

The prevention half of the reliability effort: everything in v0.30.1 *recovers* from a stalled stream; this stops the stall from happening. With the screen off, Wi-Fi power-save can throttle the radio mid-stream. ExoPlayer apps (Spotify et al.) avoid it via `setWakeMode(WAKE_MODE_NETWORK)` — a CPU WakeLock + WifiLock held while playing — but just_audio doesn't expose that API, and audio_service only holds the CPU half. This adds the missing Wi-Fi half.

## How

- **Native**: in-repo path plugin `packages/wifi_lock_shim` (~60 lines Kotlin, one `setHeld(bool)` method). A path plugin rather than a MainActivity channel so it auto-registers on **every** FlutterEngine — including the headless one Android Auto / media-resumption cold-binds boot, where an activity-registered channel would be dead. App-context lock, reference counting off (idempotent), released defensively on engine detach. Gradle mirrors audio_session 0.2.4 (already proven under this AGP 9.0.1 / Kotlin 2.3.20 toolchain).
- **Dart**: pure `shouldHoldWifiLock()` evaluated from `_broadcastState()` (fires on every play/pause/track/backend/state transition), deduped in the shim so only actual changes cross the platform channel. Fail-open, `Platform.isAndroid`-guarded.

**Held** only while the phone actively moves audio bytes over the network:
- local backend playing/buffering a network source — including a downloaded track whose file went missing (the backend silently falls back to streaming; mirrored with the same `existsSync` check), and
- a cast renderer fed **from the phone**: iroh LAN relay, local-file serve, or visualizer HLS (mirrors `resolveRendererUri`'s branching).

**Released** when paused, stopped, error-parked, playing from disk, or casting a plain-HTTP track (renderer fetches direct). A parked player never holds the lock — that would be a battery bug, and the v0.30.1 tunnel heal works without it.

## Honest coverage note

`WIFI_MODE_FULL_HIGH_PERF` (what `WAKE_MODE_NETWORK` holds) works as intended through **API 33** — which covers the Galaxy S20 FE fleet the "playback stops" reports came from (caps at Android 13). On **API 34+** the OS silently converts the acquire to `WIFI_MODE_FULL_LOW_LATENCY`, which is only active foreground-with-screen-on — i.e. inert for screen-off streaming, and no public API does better there (modern Android leans on the FGS media exemption instead). Verified against SDK sources; documented in the code so a future "still stalls on Android 15" report doesn't get misattributed to this shim.

An adversarial multi-agent review ran over the diff; its three confirmed findings (phone-origin casts beyond iroh not holding the lock, the missing-file streaming fallback classified as local, and the wrong API 34+ platform claim) are all fixed in this PR.

## Testing

- `flutter analyze` clean (pre-existing baseline only); 106/106 tests pass (6 new for `shouldHoldWifiLock`).
- Debug APK builds under the AGP 9 toolchain.
- On-device (pending): `adb shell dumpsys wifi` should show the `mstream:streaming` lock appear on play, vanish on pause/stop/disk playback, and hold while casting an iroh track; then a 30+ min screen-off streaming soak.

### DIFF
--- a/lib/media/audio_stuff.dart
+++ b/lib/media/audio_stuff.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io' show File;
 
 import 'package:audio_service/audio_service.dart';
 import 'package:audio_session/audio_session.dart'
@@ -8,6 +9,7 @@ import 'package:just_audio/just_audio.dart';
 import 'package:mstream_music/main.dart';
 import 'package:rxdart/rxdart.dart';
 import 'package:http/http.dart' as http;
+import 'package:wifi_lock_shim/wifi_lock_shim.dart';
 import 'playback_backend.dart';
 import 'local_playback_backend.dart';
 import 'dlna_playback_backend.dart';
@@ -292,6 +294,10 @@ class AudioPlayerHandler extends BaseAudioHandler
       return;
     }
     if (identical(next, _backend)) return;
+    // A visualizer cast makes the phone the renderer's byte origin (on-device
+    // HLS transcode served by LocalMediaServer) — the WifiLock heuristics need
+    // to know, and the backend itself doesn't expose it.
+    _castVisualizer = !target.isLocal && visualizer;
     await _switchBackend(next);
   }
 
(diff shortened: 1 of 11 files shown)


IrosTheBeggar:
**On-device verification (S25, wireless adb):**

- Play → `WifiLock{mstream:streaming ... workSource=mstream.music.plus}` appears in `dumpsys wifi` within seconds; acquire counter increments.
- Pause → lock released within **1s** (polled every second for 15s; stays released).
- Full acquire/release ledger balances across multiple play/pause cycles (including user-driven ones from the phone UI).
- Platform-fact confirmed live: the S25 (API 35+) counts the acquire under `full low latency` — the OS converts HIGH_PERF exactly as documented, so on 14+ the lock is held-but-inert for screen-off. The *protective* behavior this PR exists for applies to ≤ Android 13 devices (the S20 FE fleet that reported the stalls); real-world efficacy there will show up in recurrence reports, not local testing.

PR #99, opened 2026-07-08
feat(playback): survive process death via Android media resumption

IrosTheBeggar:
## What

The recovery of last resort in the reliability series: #94–#98 keep playback alive *while the process lives*; this makes an actual OS kill recoverable with one tap from Android's media-resumption chip (Quick Settings / shade, Android 11+), or a Bluetooth PLAY — instead of "find and reopen the app."

## Background

The plumbing has existed since Android Auto (v0.28.0): audio_service already answers the system's `EXTRA_RECENT` bind, and `auto_browse` already serves a playable "resume" entry. But on a **cold headless boot** (the only case that matters — the process just died) both halves raced the queue restore and lost:

1. **Recent-root race** — the reply served only the live `mediaItem` (null until QueueStore restores) *and* sat behind `ensureLoaded()`'s iroh tunnel dial (30–45s on a cold bind). System UI asks within moments; it got nothing, so no chip.
2. **Play-tap race** — a PLAY reaching the handler before the restore finished was a bare no-op on an empty player (the #97 reseed net needs a non-empty queue). This is the same defect as the audit's "Bluetooth play answered before restore" finding — fixed together.

## How

- **Chip content**: the `recent` root is answered **before** the `ensureLoaded` await, falling back to display metadata peeked straight from the persisted snapshot (`QueueStore.peekResumeItem()` — pure parse of queue.json, no server resolution, no network; unit-tested statics). The restore chain is kicked in the background. The branch also moved above the no-server notice (a local-files queue has no server, and resuming needs none).
- **Play gate**: `AudioPlayerHandler.play()` on an empty queue with the restore unsettled now (a) self-triggers `ensureLoaded → QueueStore.init` — headless boots never run main.dart's `initState` trigger — and (b) awaits a restore-settled completer (signalled from `QueueStore.init`'s `finally`: restored / nothing to restore / failed all settle it), bounded at 60s. After the wait it falls through to the existing bare-play/reseed logic, and a pause/stop that landed *during* the wait wins over the stale play.

Hardened from the adversarial review (2 lenses, 13 findings, 3 confirmed + 1 nit — all fixed):
- The chip reply no longer waits behind the tunnel dial (the original sin of my first cut).
- A pause/stop during the restore wait is honored (`_playIntent` re-check).
- An empty-queue save deletes the snapshot **only after a queue actually existed this session** — the listener attach replays the empty queue after a whiffed restore (e.g. transient server-list read failure), which used to destroy the very snapshot the chip depends on. Pre-existing bug, feature-critical now.
- The Auto resume tap no longer awaits `play()`'s never-completing future.

## Testing

- `flutter analyze` clean (baseline only); 107/107 tests pass (7 new: snapshot peek parsing, index clamping, corrupt-shape handling).
- On-device (next): kill the process, send `KEYCODE_MEDIA_PLAY` → expect headless boot → restore → resume at position; plus chip presence in the shade after death, and normal-launch regressions.
- Chip persistence after process death is ultimately at the OS/OEM's discretion (One UI has its own media-panel behavior) — this PR makes the app answer correctly whenever the system asks.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -59,8 +59,12 @@
     <!-- allowBackup=false: servers.json holds credentials (JWT, password, and the
          iroh pairing code's connect secret), so keep it out of adb / cloud backups
          where another process could exfiltrate it. -->
+    <!-- MainApplication (not the Flutter default ${applicationName}):
+         process-wide native init (IrohNative/ndk_context) that must also run
+         on HEADLESS service binds — media-resumption chip, Bluetooth media
+         key, Android Auto — where MainActivity never exists. -->
     <application
-        android:name="${applicationName}"
+        android:name=".MainApplication"
         android:label="mStream Music"
         android:allowBackup="false"
         android:usesCleartextTraffic="true"
--- a/android/app/src/main/kotlin/com/example/mstream_music/MainApplication.kt
+++ b/android/app/src/main/kotlin/com/example/mstream_music/MainApplication.kt
@@ -1,0 +1,23 @@
+package com.example.mstream_music
+
+import android.app.Application
+
+// Process-wide native init. Runs on EVERY process start — including the
+// HEADLESS service binds (media-resumption chip, Bluetooth media key, Android
+// Auto) where MainActivity never exists. IrohNative used to be initialized
+// only from MainActivity.configureFlutterEngine, so a headless boot had no
+// working tunnel: ndk_context panicked ("android context was not
+// initialized"), every iroh stream URL failed, and a saved iroh queue could
+// not resume. MainActivity keeps its own (idempotent) call as a belt.
+class MainApplication : Application() {
+    override fun onCreate() {
+        super.onCreate()
+        // Guarded so a missing/unloadable .so never blocks boot — the feature
+        // just stays unavailable (mirrors MainActivity).
+        try {
+            IrohNative.ensureInit(this)
+        } catch (e: Throwable) {
+            android.util.Log.w("IrohNative", "iroh native init failed: ${e.message}")
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


IrosTheBeggar:
**On-device verification (S25, wireless adb) — the death-and-resurrection path PASSES:**

1. Play an iroh queue → stop → `am stop-app` (OOM-like kill, process confirmed dead).
2. Bare `KEYCODE_MEDIA_PLAY` (the Bluetooth-resume scenario).
3. New process boots **headless**, the play gate triggers + awaits the restore, and playback starts with audio rendering (~10s end-to-end, verified via `dumpsys media_session` PLAYING + `dumpsys audio` started track).

The first run of this test found a real pre-existing bug now fixed in this PR (46ba8ef): **the iroh native context was initialized only from MainActivity**, so any headless boot panicked (`ndk_context: android context was not initialized`) and iroh playback was impossible — which also means Android Auto cold-binds of iroh queues have been silently broken since Auto shipped. Init moved to `MainApplication.onCreate` (every process start, headless included).

Chip rendering in the One UI quick panel is a visual check still pending (Samsung's media panel setting dependent); the underlying recent-root answer is verified serving from the snapshot.

IrosTheBeggar:
User-confirmed on One UI: the media chip persists in the quick panel after killing the app with music paused — the last open verification item. Full matrix now covered: chip content ✓, chip persistence ✓, tap/media-key resurrection ✓, headless iroh tunnel ✓.

IrosTheBeggar:
**Chip-tap resurrection captured and verified end-to-end (screenshot-driven UI testing):** with the process dead (`am stop-app`, pid empty), One UI offers a **"Play last song"** card in the Quick Settings media area. Tapping it: headless boot → the PLAY arrived 1.2s after boot, *before* the restore — the restore-settled gate held it → queue restored to the correct track → **playing at the saved position ~5s after the tap**. The full designed sequence, driven through the real OS surface rather than a synthesized media key.

PR #101, opened 2026-07-08
feat(downloads): opt-in "Keep queue available offline" auto-download

IrosTheBeggar:
## What

Part B of offline resilience, **stacked on #100** (merge that first; GitHub will retarget this to master when the base branch is deleted). An opt-in Settings toggle — default **off** — that automatically downloads every queued server track, with a Wi-Fi-only companion toggle (default **on**).

## How

- **Settings**: `offlineQueue` + `offlineQueueWifiOnly` in SettingsManager (persisted, reset-aware), two tiles under Playback (the Wi-Fi tile is disabled until the feature is on), l10n in all 10 locales.
- **Sweep**: the handler's existing queue listener (headless-safe — Android Auto sessions and Auto-DJ top-ups sweep too) calls `DownloadManager.autoDownloadQueue` on every queue emission. Selection is a pure, unit-tested function: server tracks whose file isn't on disk, at most once per session each. Candidates go through the existing `downloadOneFile` machinery — on-disk short-circuit, in-flight dedupe, FAT-name guard, iroh URL re-resolve — with `requiresWiFi` mapped to background_downloader, so the **OS** holds transfers until Wi-Fi rather than the app polling connectivity.
- **Delivery**: completed downloads flip queued copies to disk via #100's patch hook. Files land in the normal per-server download store — **no eviction by design** (documented); they're visible and deletable like any manual download.
- Flipping the toggle ON sweeps the current queue immediately; OFF cancels Wi-Fi-held tasks (running transfers finish).

## Review hardening (2 lenses, 14 verify agents — 6 confirmed findings, all fixed)

1. **Restart duplicates (the big one)**: Wi-Fi-held tasks persist in WorkManager across process death; the in-memory dedupe guards didn't, so every relaunch re-enqueued a full duplicate set (N× data). `initDownloader` now **rehydrates surviving tasks** into the guards and trackers — which also means their completions patch the queue and they show on the Downloads screen after a restart.
2. **Stranded failures**: terminal failures un-mark the once-per-session guard, and the connectivity-regained / tunnel-reconnected recovery hooks re-sweep — retries happen exactly when they can succeed, no app restart needed.
3. **Dead localPath exemption**: a queued copy whose downloaded file was deleted (SD ejected, folder cleaned) silently streamed forever while the sweep skipped it; candidates now probe `existsSync`, mirroring every playback consumer of `localPath`.
4. **Toast spam**: the sweep passes `quiet` (no per-file snackbars for an action the user didn't take; a dead server × 300-track queue used to mean ~20 minutes of sequential toasts) and the terminal-failure warn is throttled like the existing FAT-skip warn.
5. **Disable does nothing**: turning the setting off now cancels the OS-held tasks that would otherwise fire hours later.
6. **Frozen 0% rows**: Wi-Fi-held tasks show "Waiting for Wi-Fi" on the Downloads screen.

Also: `initDownloader` moved from the widget `initState` to `_startApp` so headless boots track completions (and can't double-subscribe).

## Testing

`flutter analyze` clean (baseline only); 126/126 tests pass (6 candidate-selection tests: convergence, dead-file re-candidacy, un-mark retry, duplicate collapse). APK builds. On-device verify: enable the toggle over a queue → downloads appear and the queue badges light up as they land; toggle off on cellular → held tasks vanish; kill/relaunch mid-held → no duplicate tasks (`FileDownloader().allTasks()` count stable).


IrosTheBeggar:
**On-device smoke test (S25, wireless adb) — PASS end-to-end:**

1. **Sweep**: enabled `offlineQueue` (via settings.json, headless) → relaunch → the queue-restore emission triggered the sweep → **52/52 tracks downloaded** off the LAN iroh server in ~75s, zero failures.
2. **Patch-on-complete (#100)**: one `download landed — queued copies of <track> marked local` per completion, all 52.
3. **Persistence**: `queue.json` now carries 52 `localPath` entries.
4. **Offline playback**: cold restart → restored queue plays and skips across tracks with **zero WifiLock acquires the whole session** (ledger stable at 147/147) — the #98 heuristics hold the lock for any network stream, so no-lock-while-playing is the from-disk proof. Every loaded source came off disk.
5. **Convergence**: the relaunch sweep enqueued nothing (on-disk short-circuit) — no duplicate transfers.

Environmental note, not an app bug: after a `force-stop`, `cmd media_session dispatch play` goes nowhere ("Media button session is null") until something re-registers routing; a raw `input keyevent 126` works.

IrosTheBeggar:
**UI-driven smoke round (screenshot + tap automation) — PASS:**

- Settings tiles round-trip through the real UI: toggling "Keep queue available offline" OFF greys/disables the Wi-Fi tile and persists `false`; toggling back ON persists `true` and the re-sweep converges with zero download activity (everything already on disk).
- The queue list shows the downloaded badge (green `VelvetColors.success` edge) on all 52 rows — #100's live patch made them appear without any queue rebuild.
- Post-kill "Play last song" chip → tap → headless boot → gated play → **playing from disk at the saved spot in ~5s** (also recorded on #99).

PR #103, opened 2026-07-09
fix(playback): audit stragglers — skip intent, Auto-DJ duration/auth, failed-restore spot

IrosTheBeggar:
## What

The four remaining small findings from the playback-stops audit, in one cleanup PR:

1. **Skip honors pause** — `_skipFailedTrack`'s post-skip resume is now gated on `_playIntent`: a preloading track can fail while you're deliberately paused, and the skip used to start audio over the pause. Found-by-review companion: `_switchBackend` now syncs `_playIntent` to the carried playing state, because playback started from a renderer's own remote never passes through `handler.play()` — without the sync, every intent-gated resume would land paused mid-listen after returning to the phone.
2. **Auto-DJ tracks carry `duration`** — browse parity; without it the restore clamp (#87) couldn't detect an at-the-end saved position, so reopening seeked past the end → completed → stopped ("playback won't start").
3. **Auto-DJ auth death is no longer silent** — a 401/403 logs and toasts once per expiry episode (re-arms on a successful pick and on a DJ server switch, so a *second* expiry after re-login isn't silent either — review catch); other failures log for Diagnostics.
4. **A failed launch restore keeps your place** — offline/dead-tunnel cold starts throw during the restore's load *after* the queue is published; the player then reads track 1 / 0:00 and (worse) the next background snapshot save persisted that loss. The intended spot now parks on the handler: revive paths load there, the now-playing UI shows the parked track, and snapshot saves persist it (clamped against the spot track's own duration). Cleared the moment any track actually loads or the user navigates/edits — explicit actions supersede the park.

## Review hardening

The adversarial review (2 lenses, 14 verify agents) confirmed six defects in my first cut, all fixed here: the stale spot hijacking explicit track selections made while parked (now cleared on skip/seek/edit/clear); the now-playing fallback being dead code (just_audio's `currentIndex` reads 0, never null, after a failed load — the spot must win while parked); the tunnel-heal gate keying on queue[0] instead of the parked track (a mixed queue would never heal); unclamped spot indices after queue edits (media3 throws on an out-of-range `initialIndex` and silently no-ops the seek); the once-per-session auth flag never re-arming; and the cast-remote play-intent gap above.

## Testing

`flutter analyze` clean (baseline only); 126/126 tests pass. S2/S3 are best exercised by an expired-JWT server and a long Auto-DJ session; S4's offline cold start reproduces by killing the server and restarting the app (previously landed on track 1 at 0:00 — the exact loss observed during the #99 on-device test).

### DIFF
--- a/lib/media/audio_stuff.dart
+++ b/lib/media/audio_stuff.dart
@@ -86,6 +86,10 @@ class AudioPlayerHandler extends BaseAudioHandler
 
   dynamic jsonAutoDJIgnoreList;
 
+  // One toast per session for an Auto-DJ auth failure (401/403) — every
+  // queue-end retriggers autoDJ, and each would re-toast otherwise.
+  bool _autoDJAuthWarned = false;
+
   // Session-only: the Camelot anchor for harmonic mixing. Locked on
   // the first DJ-picked song with a recognised key (after that, every
   // subsequent pick uses the anchor's 6 Camelot neighbours, keeping
@@ -226,6 +230,9 @@ class AudioPlayerHandler extends BaseAudioHandler
         _failedSkips = 0;
         _httpRetries = 0;
         _networkStalled = false;
+        // A track actually loaded — the live player state is the truth now;
+        // stop overriding reads with the failed-restore park.
+        _restoreSpot = null;
       }
       // Any fresh load re-resolves its source from the current extras
       // (_uriFor), so the loaded-source-predates-download-patch marker is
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
**On-device smoke test (S25, UI-driven) — S4 verified end-to-end with a forced restore failure:**

Planted a queue.json whose first track points at a nonexistent server path (the restore preloads item 0, so its load throws — the exact offline/dead-source cold-start shape), saved index 21 @ 42.5s. Result:

1. `[queue] restore load failed — spot parked (track 22 @ 42s)` — the park caught the failure.
2. The existing HTTP-retry machinery then consumed the spot automatically and re-seeded **onto track 22 from disk** — no manual play needed.
3. Session parked PAUSED at exactly 42468ms, item 21; **the post-restore snapshot save preserved index 21 / positionMs 42468** (pre-fix: track-1/0:00 clobber, the loss observed live during the #99 test).
4. The player bar shows "Mystery Song ft. BEGINNERS · 00:42/02:55" — the spot-first now-playing emit working (screenshot-verified; pre-fix showed track 1).
5. Cleanup + clean relaunch: zero errors, play/pause at the right spot.

S1/S2/S3 are review- and reasoning-verified (their triggers — cast-remote-started playback, a long Auto-DJ session, an expired JWT — aren't reachable in this rig); the S3 toast and S2 duration will show themselves in normal use.

PR #104, opened 2026-07-09
feat(add-server): Quick Connect discovers iroh servers on the LAN (mDNS)

IrosTheBeggar:
## What

The rescued mDNS Quick Connect work (built June 28, sat uncommitted across three worktrees), rebased onto post-v0.31.0 master and re-reviewed. Design: mDNS advertises *whether* iroh is available; the Quick Connect tab shows an "On your network" list of those servers; tapping one does the whole dance — LAN login → fetch the pairing code over the authenticated connection → start the tunnel → save as an iroh server that then **roams on cellular**. The pairing code/secret never rides the mDNS broadcast.

**Server counterparts** (also rescued + pushed tonight): mStream [#720](https://github.com/IrosTheBeggar/mStream/pull/720) (advertise `iroh=1` — ⚠️ stacked on #641, retarget before merging) and [#721](https://github.com/IrosTheBeggar/mStream/pull/721) (authed admin can fetch the code without `shareCodePublic`). Until a server runs those, the section just shows its searching state — the app side is safe to merge independently.

## Review hardening (2 lenses, 16 verify agents — 7 confirmed findings, all fixed)

1. **(high)** The discovered flow and the manual paste/scan Test ran concurrently against the **singleton** native tunnel — the second dial returns the *first* dial's port, and everything downstream (HTTP proof, save, `registerActiveTunnel`) silently binds the wrong server. Now mutually exclusive (cross-disabled UI + entry guards), and failure-path tunnel stops only fire for a dial the failing flow actually made.
2. bonsoir surfaces native NSD failures as stream *errors*; unhandled, they left discovery dead while the UI spun "Searching…" forever. Now logged and reset so Refresh genuinely restarts.
3. A stale advert (server left, IP changed) produced a login sheet doomed to fail; a connect-failure now shows "unreachable" (which also un-deads the `lanUnreachable` string).
4. `ServiceUpdated` events (DHCP moves, late TXT records on Android 14+) were dropped → stale tiles; now upserted/removed.
5. IPv6-only hosts produced unparseable URLs (unbracketed, `%zone` suffix); now formatted defensively.
6. **(security)** The https bootstrap added blanket self-signed trust for any tapped host and never removed it; the pending trust is now scoped to the bootstrap and dropped in the `finally`.
7. Mounted guards after the network awaits (login sheet on a defunct context).

Accepted/documented: the empty state is a spinner + Refresh (no "none found" timeout), and a wrong password reports in the shared result banner rather than re-opening the sheet.

## Testing

`flutter analyze` clean (baseline only); 126/126 tests; APK builds and is deployed to the test device. Full end-to-end (discover → tap → roam) needs a LAN server running #720+#641 — pending those merges + a server deploy; the discovery/UI layer degrades to the searching state until then.


IrosTheBeggar:
**Full end-to-end smoke test — PASS.** Stood up a real mStream server locally (merged master + the #641/#720/#721 branches into a throwaway stack, iroh + mDNS + code-sharing all on, public mode, mdns name "QC Smoke"), deployed the app to the S25, and drove Quick Connect through the actual UI:

1. **Discovery**: the "QC Smoke — mStream v6.17.2" tile appeared under "On your network" — mDNS browse, the `iroh=1` TXT filter, and version parsing all correct. (A non-iroh server would not appear; the existing demo HTTP server didn't.)
2. **Bootstrap**: tap → ping → public-mode detected (no login) → pairing code fetched over LAN → tunnel dialed → HTTP proof → **saved as `iroh://5e7334b` with a green "Direct" badge**, public-access auto-toggled ON. The mutual-exclusion fix was visible live: the manual Test/Scan/Paste controls greyed out while the tile spun.

**One real bug the smoke test caught and fixed** (new commit): my dev host has a WSL virtual adapter, so the server advertised `172.28.0.1` (unroutable from the phone) *first*, alongside its real `192.168.1.71`. The old `hostAddresses.first` dead-ended on "Couldn't reach this server". Now `DiscoveredServer.baseUrls` exposes every advertised address and the connect flow probes each until one answers — verified reaching the real LAN IP past the WSL one. This is the common case for anyone whose server host runs WSL/Docker/VPN/Hyper-V, so a genuinely valuable catch.

Analyzer clean, 126/126 tests. The server counterparts (#720/#721) are validated by this same run.

IrosTheBeggar:
Paul's manual login-flow test surfaced two bugs; both root-caused on-device and fixed in the latest commit.

**1. Tile shows immediately OR never (until Refresh).** Not an mDNS problem at all — wire sniffing showed queries answered in 20ms, and instrumentation proved the resolved service (full TXT, addresses) reached the Dart singleton every time. The tile's StreamBuilder just subscribed too late: discovery starts when the screen opens, the Quick Connect tab builds on tab-switch, and a resolve landing in between was emitted with no listener — broadcast streams don't replay. Fix: the `stream` getter now yields the current snapshot to each new subscriber before live events. Also hardened while in there: bonsoir's `ResolveFailed` (previously ignored — the event carries no service upstream, so retry works off our own pending set) retries with backoff; a watchdog auto-restarts discovery if the platform resolve wedges (Android NSD can hang, and bonsoir's resolve queue only drains on a callback); discovery events now appLog for future diagnosis.

**2. Full-screen error page after signing in.** The login sheet disposed its `TextEditingController`s in `whenComplete()` — which fires when the pop *starts*, while the closing sheet keeps rebuilding through its exit animation (the collapsing keyboard changes viewInsets). TextField + disposed controller → build exception → ErrorWidget over the still-running flow (which is why back-button landed in a working app with the server saved). The sheet is now a StatefulWidget owning its controllers; same latent bug fixed in the new-folder dialog.

Verified on-device: cold entry → tile in seconds, no refresh; sign-in → sheet closes clean, no error page, server saved with Direct badge. 126/126 tests, analyzer clean.

IrosTheBeggar:
One more from Paul's testing: leaving the Quick Connect tab and returning threw a full-screen **'Stream has already been listened to'**. My snapshot-replay fix had used an `async*` generator — single-subscription — and the tab's StreamBuilder widget re-listens to the *same* stream object when TabBarView remounts the tab (the widget instances are created once per screen build). Swapped to `Stream.multi`, which keeps the per-subscriber snapshot replay but tolerates any number of (re)listens. Verified on-device with the exact repro: QC → Server URL → QC shows the tile cleanly. 126/126 tests.

PR #107, opened 2026-07-12
iOS support: full port + iroh Quick Connect

IrosTheBeggar:
## Summary

Makes the app build and run on iOS, including iroh Quick Connect. Two commits:

**`941571a` — feat(ios): enable the iOS build**
- Regenerated `ios/` with the Flutter 3.44 scaffold; **all 15 plugins resolve via Swift Package Manager** — no Podfile/CocoaPods. Bundle id `mstream.music`, deployment target 15.0 (flutter_chrome_cast's floor; background_downloader/media_cast_dlna need 14).
- Info.plist: background audio, ATS exception (mStream servers are commonly plain `http://`), camera (QR pairing), `NSBonjourServices _mstream._tcp` + local-network usage, mic description.
- Android-only UI hidden on iOS: EQ / visualizer / imported-shaders / cast-quality settings tiles and the storage-mode picker (iOS always downloads into the app sandbox); `FileExplorer.getDownloadDir` resolves legacy Android storage modes to the documents dir off-Android.
- `flutter_launcher_icons`: `remove_alpha_ios` (the App Store rejects icons with an alpha channel).
- Integration-test fixes that help every platform: the MockServer now serves HTTP byte ranges (AVPlayer refuses no-range servers with error -11850; ExoPlayer doesn't care) and the playback test uses a 30s WAV so assertions sample mid-playback.

**`c75abec` — feat(iroh): Quick Connect on iOS**
- `rust/iroh_tunnel` builds for `aarch64-apple-ios` with zero source changes (Android code was already cfg-gated; deps are ring + noq, both iOS-clean).
- New `rust/iroh_tunnel/build-ios.sh` produces a committed 14 MB `iroh_tunnel.xcframework` (same committed-binary model as the Android jniLibs `.so` — re-run + commit after touching the crate).
- `packages/iroh_tunnel_native`: in-repo SwiftPM ffi plugin vending the xcframework; Flutter links it into Runner and Xcode embeds + signs it automatically.
- `lib/native/iroh_tunnel.dart`: platform-aware loader; `isSupported` is now true on iOS, and the entire Quick Connect surface (tab, QR scan, LAN discovery, tunnel lifecycle, status banner, reconnect/auto-heal) chains off it with no other Dart changes. Desktop still returns false.

## Verified
- Release device build and simulator build both link + embed everything (framework signed at `Runner.app/Frameworks/`)
- App launches on the iOS 26.5 simulator; notification permission flows work
- Integration tests on the simulator: add-server ✅, playback-progress ✅ (real HTTP streaming through AVPlayer)
- FFI smoke test: dlopen of the embedded framework, all 9 `mstream_iroh_*` symbols, live status/pathKind/localToken calls into the Rust runtime
- `flutter analyze` clean (only the 17 pre-existing infos), all 131 unit tests pass

## Not covered yet
- Signing/provisioning (no Apple Developer team configured) and physical-device testing: lock-screen controls, background audio, AirPlay routing, tunnel recovery after background suspension
- E2E iroh pairing against a PR mstream#643-enabled server (`rust/iroh_tunnel/interop/harness.mjs` can stand one up); QR scanning needs a real camera
- The remaining 10 integration tests on iOS; App Store chores (privacy questionnaire, ATS justification, export compliance)

## Platform notes
- DLNA casting stays Android-only (upstream dropped iOS support); Chromecast compiles in but isn't registered on iOS — AirPlay covers casting via the system audio route. Visualizer/EQ remain Android-only behind their existing guards.
- Merge overlap heads-up: `feat/windows-desktop` also touches `pubspec.yaml` / `.metadata` / `test_helpers.dart`; whichever branch lands second needs a small rebase.


IrosTheBeggar:
Real-device testing on an iPhone X (iOS 15.6) caught two shipping bugs in this PR; both fixed in the two new commits:

- **dbbbbc9** — the app crashed at launch on every iOS 15–17 device: the `netdev` crate (iroh transitive dep) hard-links `nw_path_is_ultra_constrained`, an iOS 18-only Network.framework symbol, and dyld aborts on the missing import. Simulators never caught it (they run iOS 18+). Fixed with a forwarding shim inside the dylib (`rust/iroh_tunnel/src/apple_compat.rs`) — real implementation via dlsym on iOS 18+, graceful fallback below.
- **cf1e1c4** — the shader visualizer had no exit on iOS (desktop closes via window chrome, Android via system back; the tap gesture cycles presets) and `MultiPassRenderer` leaked one undisposed `Picture` per buffer pass per frame, pinning full-size feedback buffers until iOS jetsam-killed the app (`per-process-limit(7)`, no crash report) after a few minutes of Cyber Fuji / Hex Marching. Added a close button (notch/home-indicator safe) and disposal.

Verified on the device: launch, playback, visualizer soak (10 min multi-pass, no kill). The visualizer files here intentionally match what feat/windows-desktop will want — the picture leak exists there too.

IrosTheBeggar:
One more device finding, fixed in **2333dd0**: downloads were a silent no-op on iOS for any file path containing a space (i.e. most real libraries). Download URLs are concatenated raw, Android's stack tolerates the unencoded result, but iOS's `URL(string:)` rejects the task — and `enqueue()`'s `false` return was ignored, so nothing surfaced. Fixed by normalizing the URL through Dart's `Uri` in `downloadOneFile` (idempotent; same treatment stream URLs already get) and surfacing refused enqueues. Device-verified: spacey-path tracks download and appear in the local file browser.

IrosTheBeggar:
Two more device-verified fixes in **98d06c8**:
- **Deleting a server now removes its songs from the queue** (platform-neutral — Android had the same gap). Pure, unit-tested queue-filter helper + the established reload choreography; Auto-DJ gets unpointed so it can't refill from a deleted server.
- **iroh tunnel recovery after iOS suspension**: pausing and backgrounding for a few minutes kills the QUIC connection, but on thaw the native status still reads *connected*, so the resume-time verify never rebuilt. The app now probes ground truth on resume (one tokened HTTP round-trip through the loopback on a fresh connection) and hard-rebuilds the tunnel when the path is dead.

PR #109, opened 2026-07-12
iOS visualizer: real audio reactivity (Rust decode sidecar)

IrosTheBeggar:
Phase 2 of the iOS visualizer (stacked on #107, which ships the shader engine with a synthesized signal): the visualizer now reacts to the track that's actually playing.

## How

- **`rust/viz_decoder`** — a small sidecar crate that re-decodes the currently-playing source with [symphonia](https://crates.io/crates/symphonia) into a rolling ring of mono PCM, chasing the playback position on a background thread (decode-ahead ~4 s, coarse seek on scrubs, EOF-clamped reads). Sources: downloaded files, and HTTP(S) with a Range-backed seekable reader (lazy seeks, one transparent mid-stream reconnect) that also handles no-Range endpoints (transcode) by forward-chasing. The C ABI (`mstream_vizdec_*`, 7 symbols) is panic-guarded like `iroh_tunnel`'s; `read` is non-blocking and serves the exact window ending at a position, so it's safe at frame rate on the UI isolate.
- **Dart keeps the FFT.** `SpectrumSource` now prefers the sidecar window (keyed to a position extrapolated from `PlaybackState.position/updateTime/speed`), then the Windows WASAPI capture, then synth. `ShaderVisualizerScreen` starts the sidecar from `AudioPlayerHandler.currentPlayableUri()` (new: downloaded file, else the tunnel/transcode-fresh stream URL) and retargets it on every track change.
- **Packaging mirrors `iroh_tunnel` exactly**: committed `viz_decoder.xcframework` (5.9 MB, device+sim arm64) built by `rust/viz_decoder/build-ios.sh`, vended to Runner by the `viz_decoder_native` SwiftPM ffi plugin.

## Fail-open by design

Any failure — unreachable server, self-signed HTTPS cert (rustls + webpki roots; verification is NOT disabled), unsupported codec — kills the session quietly and the visualizer falls back to the synthesized signal. Notably: **opus transcodes stay synthesized** (symphonia has no opus decoder); mp3/aac transcodes and all direct formats (flac, mp3, m4a/alac, ogg, wav, aiff) decode fine.

## Trade-off worth knowing

For streamed (non-downloaded) tracks the sidecar opens its own connection to the same URL, so a track streams twice while the visualizer is open (bounded by the ~4 s decode-ahead; the ring holds ~12 s). Battery/bandwidth cost is visualizer-open-only.

## Verification

- `cargo test` (host): 6 e2e tests — sample-exact windows through file reads, forward/backward scrubs, HTTP Range, a 200-only server chase, EOF clamp, and fail-open on a dead source. `cargo clippy` clean.
- iOS 26.5 simulator: an FFI e2e against the real embedded framework passed file + HTTP Range decode, both scrub directions with sample-exact asserts, the `SpectrumSource` seam (a 440 Hz sine produced the single dominant FFT bin ≈ 10, which synth never does), and fail-open flipping the Dart wrapper off.
- `flutter analyze` clean (baseline), 131 unit tests pass, `flutter build ios --simulator` embeds + signs the framework.

Not covered yet (needs a device / follow-ups): physical-device perf + battery, background-suspension behavior, and desktop adopting this crate to replace the Windows-only WASAPI shim.


IrosTheBeggar:
Rebased onto the updated base (launch-crash shim + visualizer exit/leak fixes now live in #107) and added **e465572** — the real-device tuning round from an iPhone X (iOS 15.6):

- **Spectrum auto-gain** for real-audio frames (running-peak normalize, instant attack / ~2 s decay, gain capped 42×): real music spreads energy across the spectrum, so the bars barely registered against shaders tuned on the synthetic carriers. Also benefits the Windows loopback path.
- **Landscape lock** while the visualizer is open (presets are wide Shadertoy compositions); restores the app default on close.
- **~9× less fragment work on phones**: passes render offscreen at logical resolution and GPU-upscale instead of shading every device pixel (2.7M invocations/pass at 3× DPR; Hex Marching stacks three). Desktop keeps the original full-res path.
- Every snapshot `Picture` disposed immediately (the offscreen path would otherwise amplify the jetsam-grade leak fixed in #107).

Device-verified end to end: real audio reactivity against a live mStream server, scrubs, track changes, orientation in/out, and a 10-minute heavy-preset soak with no memory kill.

PR #112, opened 2026-07-15
iOS: auto-transcode formats AVPlayer can't play (ogg/opus & co)

IrosTheBeggar:
Closes the last known product gap before App Store submission: **ogg/vorbis, opus, wma, and a tail of rarer formats silently fail on iOS** (AVPlayer can't decode them; Android's ExoPlayer and desktop's media_kit can).

## How

`buildServerStreamUrl` — the single funnel every playback path uses (browse, queue restore, add-folder, AutoDJ) — now routes those paths through `/transcode` **on iOS only**, even with the user's transcode setting off. Two pure, host-unit-tested helpers make the policy explicit:

- `needsIosTranscodeFallback(path, isIOS, transcodeAvailable)` — extension-based detection (`ogg/oga/opus/spx/wma/ape/mpc/wv/tta/shn/mka/dsf/dff`), honoring the existing optimistic-until-pinged `transcodeAvailable` policy; servers confirmed without ffmpeg opt out.
- `effectiveTranscodeCodec(userCodec, isIOS)` — iOS pins everything except an explicit `aac` to `mp3`: an opus transcode (chosen or server-default) would swap one unplayable stream for another. Off-iOS, user choice passes through untouched.

## Android/desktop impact: none

The gate is `Platform.isIOS` at the call site; off-iOS every URL is byte-identical to before (covered by tests).

## Verification

- **Go/no-go first**: simulator test proved AVPlayer plays the `/transcode` transport shape — chunked, no `Range`, no `Content-Length`, ADTS-AAC — to completion (20 s stream, position reached the end). The seekbar duration comes from server metadata in the MediaItem, so the unsized stream doesn't break the UI.
- On-simulator test of the real builder: `.ogg` → `/transcode` + `codec=mp3`, `.mp3` → `/media`, `transcodeAvailable=false` → no fallback, `opus` user codec → pinned `mp3`, `aac` preserved.
- 13 new host unit tests; 144 total pass; analyzer at baseline.

## Known gap (deliberate)

**Downloaded** copies of these formats still play from the raw local file and stay unplayable on iOS — this PR covers streaming. Downloading transcoded copies on iOS is a scoped follow-up.

### DIFF
--- a/lib/media/local_playback_backend.dart
+++ b/lib/media/local_playback_backend.dart
@@ -3,6 +3,7 @@ import 'dart:io' show File, Platform;
 
 import 'package:audio_service/audio_service.dart' show MediaItem;
 import 'package:just_audio/just_audio.dart';
+import 'package:rxdart/rxdart.dart';
 
 import 'playback_backend.dart';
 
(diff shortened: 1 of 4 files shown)


IrosTheBeggar:
Follow-up pushed: `e1edfba` fixes the frozen-elapsed wart found while smoke-testing. AVPlayer reports the chunked /transcode stream's duration as **zero** (not null), which just_audio treats as a real duration and clamps every position to — so the track played pinned at 00:00. The backend now normalizes zero→null (UI falls into its existing elapsed-only, seek-disabled mode) and emits unclamped extrapolated positions while the duration reads zero. Sized streams untouched. Sim-verified: ogg fallback ticks 00:18→00:47; restored mp3 still shows elapsed/total. +6 unit tests (369 total green).

PR #115, opened 2026-07-28
feat: key native iroh tunnels by pairing code (ABI v2) — multi-iroh groundwork

IrosTheBeggar:
PR 1 of 2 toward removing the one-iroh-server limit. This one changes only the native layer and its FFI plumbing; **app behavior is identical** — the one-tunnel-at-a-time policy (and the one-iroh-server cap in the add-server UI) still stand and now live purely in `ServerManager`. PR 2 will replace that policy with per-server tunnel reconciliation and drop the UI cap.

## What changed

**Rust (`rust/iroh_tunnel`)**
- `ffi.rs`: the single global `Mutex<Option<Tunnel>>` slot is now a `Mutex<BTreeMap<String, Tunnel>>` keyed by pairing code. `start` is idempotent per code (concurrent dials for the same code keep the winner, the loser tears down in the background — same race handling as before, per key). Distinct codes can run concurrently, each with its own loopback port, token, and reconnect supervisor — the `Tunnel` core in `lib.rs` was already fully instance-based and is untouched.
- `c_api.rs`: every per-tunnel entry point (`stop` / `status` / `path_kind` / `is_active` / `local_token`) takes the pairing code; `network_changed` nudges **all** running tunnels. New `mstream_iroh_abi_version` symbol returns 2.
- Dev CLI (`src/bin/client.rs`) updated to the keyed calls.

**Dart**
- `lib/native/iroh_tunnel.dart`: per-code API (`stop(code)`, `statusOf(code)`, `pathKindOf(code)`, `localTokenOf(code)`), plus an `abiVersion` getter that probes the version symbol (absent → 1). Dropped the unused `isActive` getter.
- `lib/singletons/server_list.dart` / `lib/screens/add_server.dart`: pass the code at every call site. The add-server screen now records the exact code it dialed for the test tunnel (`_irohTestedCode`) and stops by that, instead of stopping "the" global tunnel — the text field can be edited after a test.

## iOS compatibility (why the v1 symbols keep their names)

The committed `iroh_tunnel.xcframework` can only be rebuilt on a Mac (`build-ios.sh`), so ABI v2 **appends** the pairing-code parameter to the existing symbol names rather than renaming them. Against the stale v1 framework the extra register argument is ignored (AAPCS64), leaving the old single-tunnel semantics — which is exactly how the app drives the tunnel in this PR, so iOS keeps working unchanged. The app can detect v1 via the missing version symbol; PR 2 will gate "add a second iroh server" on `abiVersion >= 2`.

**⚠️ Action needed before PR 2 lands on iOS:** re-run `rust/iroh_tunnel/build-ios.sh` on a Mac and commit the updated xcframework (the script's symbol smoke-check now expects 10 exports).

## Verification

- `cargo build --release` + `cargo test` (7/7) on host
- `flutter analyze` — no new findings in touched files (18 pre-existing baseline infos elsewhere)
- `flutter test` — 137/137 pass
- Android `.so` rebuilt via `build-android.sh` for arm64-v8a + x86_64; `llvm-nm -D` confirms all 10 `mstream_iroh_*` symbols exported in both ABIs (size delta ~13 KB)


IrosTheBeggar:
Superseded: the ABI v2 keyed-tunnel work was redone on the current supervisor code and merged as #148 (with #149, #150, #151 and #152 building on it).

PR #118, opened 2026-08-17
Audit batch 3: playback/cast robustness

IrosTheBeggar:
Fixes the playback-core robustness findings from the codebase audit — 5 medium-severity issues (#23, #24, #16, #14, #18 in the audit report) plus the adjacent one-line lows in the same files (#25–#29, minus the shuffle-semantics redesign).

> **Stacked PR** — based on `claude/audit-batch2-http` (#121). Merge order: #119 (high-severity) → #120 (batch 1) → #121 (batch 2) → this.

## Fixes

**Cast teardown can no longer wedge the switch chain.** `_endSessionQuietly` timeboxes the Cast SDK call (10s). The file's own comment documents that plugin calls against a broken session can hang forever — the re-attach path already had a 30s guard, but `disposeRenderer` (awaited on `_switchChain`, and invoked precisely when the session is broken) had none. One hang froze every future cast switch/fallback until app restart.

**Suffix Range requests serve the right bytes.** `LocalMediaServer` parsed `bytes=-N` (the last N bytes — the standard renderer tail-probe for end-of-file moov atoms / ID3v1 tags) as `0..N` and served the *head* of the file as a well-formed 206. Range parsing is now a pure, unit-tested `parseRange` that handles the suffix form, ignores invalid syntax per RFC 9110, and 416s only genuinely unsatisfiable ranges. Serving tokens are minted with `Random.secure()` (they gate LAN access on an all-interfaces bind).

**Auto DJ reentrancy.** `autoDJ()` drops overlapping runs (in-flight guard). Concurrent runs share the ignoreList — updated only after the response — so the empty-queue start path, which provably fires a third concurrent call via the index listener, could queue duplicate picks on a narrow pool.

**Server-switch races.** The app-bar switch awaits `changeCurrentServer` before its `throwErr` ping (mirroring the add-server flow) — switching to an iroh server used to reliably toast a spurious "Failed to connect" while racing the tunnel bring-up. And `goToNavScreen` now cancels in-flight browse loads, so a slow response from the *old* server can no longer push its rows onto the *new* server's home screen (tapping those rows mixed servers).

**The one-liners:** backend swaps/reseeds pass `initialIndex`/`initialPosition` so re-subscribed listeners replay the real spot instead of a spurious index-0 (now-playing flash + possible double DJ top-up); `skipToQueueItem` rejects `index == length`; DLNA `seek` returns after a failed load walk instead of yanking the substitute track to the failed track's offset; the cast-failure fallback resumes with the live `_playIntent` so a pause during the 12s start wait wins.

## Verification

- `flutter analyze` clean; **200 tests pass**, including new coverage: a thorough `parseRange` suite (suffix/open/unsatisfiable/invalid forms), a DLNA regression test that a failed seek-load walks on *without* seeking the substitute, and a browser-load test that a stack reset aborts in-flight fetches.
- iOS Simulator smoke (results in the session log): queue restore, playback, and a staged server-switch race against a deliberately stalling server.
- Needs a device pass for the cast paths (real Chromecast: teardown timeout, pause-during-start) and an iroh server for the tunnel-race toast — same list as the earlier audit branches.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -888,7 +888,11 @@ class _MStreamAppState extends State<MStreamApp> with WidgetsBindingObserver {
                   child: PopupMenuButton(
                       onSelected: (int selectedServerIndex) async {
                         if (selectedServerIndex > -1) {
-                          ServerManager()
+                          // Awaited (mirroring the add-server flow): the ping
+                          // below used to race the iroh tunnel bring-up inside
+                          // changeCurrentServer and reliably showed a spurious
+                          // "failed to connect" toast on every iroh switch.
+                          await ServerManager()
                               .changeCurrentServer(selectedServerIndex);
 
                           try {
--- a/lib/media/audio_stuff.dart
+++ b/lib/media/audio_stuff.dart
@@ -716,11 +716,14 @@ class AudioPlayerHandler extends BaseAudioHandler
     // the load so the renderer auto-plays when its media is ready. Not routed
     // through _loadAtSpot — parking a local-playback spot on a switch TO a
     // renderer would be meaningless, and the local arm's own revive paths
-    // already cover the way back.
-    await next.setSources(
-        identical(next, _localBackend) ? _queueWithFreshUrls() : queue.value,
-        initialIndex: idx,
-        initialPosition: pos);
+    // already cover the way back. Clamped: a queue edit racing the switch can
+    // leave idx past the end, and an out-of-range initialIndex kills the load.
+    final items = identical(next, _localBackend)
+        ? _queueWithFreshUrls()
+        : queue.value;
+    await next.setSources(items,
+        initialIndex: items.isEmpty ? null : idx.clamp(0, items.length - 1),
+        initialPosition: items.isEmpty ? null : pos);
     _backendSubject.add(next);
     if (queue.value.isNotEmpty) {
       // Carry the play state into the load: a renderer then auto-plays when its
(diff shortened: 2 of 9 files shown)


IrosTheBeggar:
Rebased onto master (base retargeted; single commit, was stacked on batch2). This was the delicate one — master's autoDJ() and reseed paths have been heavily reworked since Aug 17 — so each sub-fix was reconciled against current master rather than blindly merged:

**Kept (still needed):** the `_djInFlight` guard (#16) now wraps master's current `_autoDJPick` body — epoch capture and all — and closes the last known bounded Auto-DJ multi-fire from the runaway audit; cast teardown timebox (#23); parseRange extraction + suffix `bytes=-N` fix + `Random.secure()` tokens (#24/#29); awaited `changeCurrentServer` (#14); `goToNavScreen` cancels in-flight browse loads (#18); DLNA seek return after a failed walk (#27); live `_playIntent` on the cast-loss reseed (#28); plus the `_switchBackend` initialIndex clamp.

**Dropped as superseded on master:** the initialIndex/-Position hunks on the two local reseed sites (master's restore series routed those through `_loadAtSpot`, which already carries the spot — reapplying them would have double-loaded), and the `skipToQueueItem >=` bound (already on master).

356 unit tests pass (batch3 adds 7), all four integration suites green on the simulator, analyze clean. Dev build installed on the phone for the device pass.

PR #121, opened 2026-08-22
Audit batch 2: HTTP layer — timeouts, shared client, off-isolate decode, batch enqueue

IrosTheBeggar:
Batch 2 of the audit's medium-severity findings: the HTTP-layer family (#38 timeouts, #45 connection churn, #41 UI-isolate decode, #40 O(N²) enqueue, #39 serial playFromHere builds, #1 unguarded dynamic dispatch).

> **Stacked PR** — based on `claude/audit-batch1-persistence` (#120). Merge order: #119 → #120 → this → #118.

## Fixes

**Timeouts.** Every direct call is bounded to 15s (`getGenres`, `sharePlaylist`, `fetchLyrics`, `rateSong`, `fetchTrackMetadata` — matching the existing playlist-call precedent), and `makeServerCall`'s non-iroh send gets a 30s bound. Against a black-holed server these hung for the OS TCP timeout (minutes); `rateSong`'s hang also left the star UI's optimistic update showing forever, since its catch is the only revert path.

**Connection reuse.** The direct (non-browse) calls share one keep-alive `http.Client` instead of a fresh client — a TCP+TLS handshake — per call; queue enrichment posts once per track and Discover fires 3–4 per seed. The client resets on server edit/remove. `makeServerCall` keeps its per-request client on purpose: closing it is what makes Back-cancel abort.

**Off-isolate decode.** Browse responses over 100KB (whole-library artists/albums/playlist/recursive payloads, multi-MB on big libraries) decode via `compute` instead of blocking the UI thread mid-animation.

**Batch enqueue.** New `AudioHandler.addQueueItems` + backend `addSources` (just_audio `addAudioSources`; single `addAll` on the emulated cast base): one queue emission and one platform call for a whole batch. "Add all" used to fire an unawaited per-item add where every add re-ran the whole-queue listeners (iroh scan + offline sweep) — O(N²) on big folders.

**playFromHere/addRowsToQueue.** Item builds run 4-concurrent (was one serial /db/metadata round-trip per row), the download dir resolves once per server (was a platform-channel call per row), and the tapped index is remapped over failed builds instead of drifting onto the wrong track.

**Guarded parsing.** The nine browse functions' response parsing moved inside their try blocks — a response-shape surprise threw NoSuchMethodError past the guard and silently killed the browse mid-function. `getRecursiveFiles`' server param is now required (its only caller always passed it), removing the `useThisServer!` deref.

## Verification

- `flutter analyze` clean; full suite passes with new tests: batch-append semantics on the emulated backend (order, index promotion, empty no-op) and the decode helper (small/large/malformed, off-isolate result identical to inline).
- iOS Simulator smoke against the demo server: "Add all" on a folder landed all 26 tracks as **one** batch (single queue emission, snapshot valid), and a batch-added track streams and plays through the new `addAudioSources` path.
- Not separately staged: a stalling server to watch the 30s browse bound fire, and the iroh retry path (untouched logic, but adjacent) — both worth a glance in the device pass.

### DIFF
--- a/integration_test/batch_add_test.dart
+++ b/integration_test/batch_add_test.dart
@@ -1,0 +1,77 @@
+// Batch add ("Add all") onto an empty queue (PR #121 follow-up).
+//
+// The regression this pins: addQueueItems onto a fresh IDLE player let
+// just_audio's idle stub land the current index on the LAST row, so the
+// now-playing surfaces showed the final track and a later play() re-seeded
+// there too ("added a folder, and it opened on the last song"). The batch
+// add must park the queue on track 1, and playback must start there.
+
+import 'package:audio_service/audio_service.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import 'package:mstream_music/singletons/media.dart';
+
+import 'helpers/test_helpers.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  setUpAll(() async {
+    await MediaManager().start();
+  });
+
+  Future<void> waitFor(bool Function() cond, String what,
+      {Duration timeout = const Duration(seconds: 20)}) async {
+    final deadline = DateTime.now().add(timeout);
+    while (!cond()) {
+      if (DateTime.now().isAfter(deadline)) fail('timed out waiting for $what');
+      await Future<void>.delayed(const Duration(milliseconds: 100));
+    }
+  }
+
+  testWidgets('batch add onto an empty queue opens and plays from track 1',
+      (WidgetTester tester) async {
+    await resetAppState();
+    final mock = await MockServer.start(
+      const {},
+      defaultHandler: (_) => buildSilentWav(seconds: 60),
+    );
+    addTearDown(mock.close);
+
+    final handler = MediaManager().audioHandler;
+    await handler.customAction('clearPlaylist');
+
+    final items = [
+      for (var i = 1; i <= 5; i++)
+        MediaItem(
+          id: '${mock.url}/media/demo/track_$i.wav',
+          title: 'Track $i',
+          extras: {'server': 'batch-test', 'path': 'demo/track_$i.wav'},
+        ),
+    ];
+    await handler.addQueueItems(items);
+
+    expect(handler.queue.value.length, 5);
+    // The parked spot must present TRACK 1 as now-playing, not wherever the
+    // idle stub's post-append emission landed (observed: the last row).
+    await waitFor(() => handler.mediaItem.valueOrNull?.title == 'Track 1',
+        'now-playing to settle on the first track');
+
+    // And play() must start there — the idle re-seed reads the same spot.
+    await handler.play();
+    // Generous: the iOS simulator's AVPlayer probes each source serially
+    // before ready — well over the Android emulator's near-instant start.
+    await waitFor(
+        () =>
+            handler.playbackState.value.processingState ==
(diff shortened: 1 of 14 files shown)


IrosTheBeggar:
Phone smoke caught a runaway the emulator pass missed: **Add All onto a fresh-boot idle player with Auto DJ armed** looped — just_audio's idle stub emitted the last row's index after the batch add, the queue-end listener read that as "playback reached the end" and appended a DJ pick, each append re-emitted the new last index, and 11 picks stacked in 9 seconds ("skipping every few seconds"). Log forensics: `add 30 tracks (now 30)` → `track 30/30` → `track 31/31` → … → settled only when the user tapped a track and reseeded.

Fixed in 38e1881: the top-up trigger now runs through a pure `shouldTopUpAutoDJ` that also requires a real playback session (not idle) — advancing into the last track (ready) and finishing on it (completed, the infinite-play case) still top up; idle stub churn cannot. Unit-tested with the runaway's exact emission shapes; 348 tests + the AutoDJ integration suite pass.

IrosTheBeggar:
Second phone-smoke find, same idle-stub root as the DJ runaway but with the DJ off: batch add onto a fresh-boot empty queue left now-playing parked on the LAST item (the stub's landing row on Android) and did not start playback. Fixed in a0a818d: `addQueueItems` parks the failed-restore spot on track 1 when batch-adding onto an empty idle player (parked before the backend append so the stub emission's now-playing emit already reads it; an explicit tap still overwrites it, and a real load reaching ready clears it), and `getRecursiveFiles` now auto-plays on an empty queue — the contract `addRowsToQueue` already documents. Pinned by `integration_test/batch_add_test.dart` on the real handler (passes on the Android emulator and the iOS simulator), plus a UI replication of the exact report on the emulator: Add All on 42 tracks → the stub still emitted `track 42/42` → playback opened on track 1 playing.

IrosTheBeggar:
F1 guard from the runaway audit applied in b9626ab: `shouldTopUpAutoDJ` now also refuses to fire while the failed-track skip walk is live (`_skipPending` or a non-zero skip budget). This closes the last unbounded loop in the family — broken streams + healthy random-songs API used to make the walk and the DJ feed each other forever, since both of the walk's exit conditions chase the queue the DJ keeps growing. Starved, the walk terminates through its normal end-of-queue logic and a `ready` re-enables the DJ. Unit-tested; 349 tests + both AutoDJ/batch-add integration tests pass on sim and emulator. (Also fixed the batch_add test's flakiness on the iOS sim — an `await handler.play()` was serializing the test behind just_audio's play-until-ends future; production callers never await it.)

IrosTheBeggar:
96ce638: Add All now queues full-fidelity tracks — ONE batched /db/metadata/batch request per folder (version-gated, best-effort; unscanned/old-server rows stay bare and keep the lazy per-track top-up), plus the download dir resolved once so already-on-disk copies carry localPath and play from disk immediately. Verified on the emulator: queue rows show title/artist/duration and the player gets album/year + key/BPM badges straight from an Add All.

Also investigated the 'two tracks never downloaded' phone report — **not a download bug**: the log shows `add 52 tracks` then `[auto-dl] evicted 1 over cap 50`. The keep-queue-offline cap defaults to 50 and caches a window AHEAD of the playing track, so positions 51–52 sat outside it (and one stale auto-download beyond the window was evicted, per the setting's own subtitle: 'songs left behind are removed as you go'). They download as playback approaches; cap 0 caches the whole queue. The sweep now logs when the window clips queued tracks (debounced), so this stops masquerading as a bug.

PR #128, opened 2026-08-29
feat(autodj): multi-server sessions

IrosTheBeggar:
## What

Auto DJ has always played from one server. This adds a mode where **every eligible server contributes**, with each pick going to whichever answers closest to the sound of what's playing.

Needs **[mStream #929](https://github.com/IrosTheBeggar/mStream/pull/929)** — the two server APIs this is built on.

## The seed is the problem it solves

A single-server session seeds from a **filepath**, which names a row on the server that holds it and means nothing anywhere else. This one reads the *vector* behind that path (`/discovery/local/embeddings`), averages the anchor's vectors — possibly across servers, which is what makes it one session rather than several running side by side — and sends that to everyone as `similarToVector`. Each server ranks its own library and returns a cosine; the best answer wins.

## The panel, restructured

Reorganised around what a setting actually governs:

- **The switch sits at the top** — turning it on rearranges everything under it, and a control that reorganises a form belongs above the form.
- **"The session"** — sonic similarity, BPM, harmonic mixing, track length, keyword. Settings that mean the same thing on every server.
- **"Each library"** — the server dropdown, **moved here**, plus that server's own vpaths, minimum rating and genre filter.

The dropdown changes meaning with the mode, which is the point: in a single-server session it *moves the DJ*; here it only chooses whose settings are on screen, because every eligible server is already playing.

## Degrading honestly

A server can sit out for three different reasons — too old, discovery off, or a different embedding model — and a silently-missing server looks exactly like the feature not working. So the section reports how many are taking part and why the rest aren't.

A model mismatch can only be discovered by asking, so that server is dropped for the session on the server's 400 rather than re-asked every pick.

**It falls back rather than failing.** No seed, fewer than two eligible servers, or nobody answering all return to the ordinary single-server pick — switching this on can never leave the queue empty where it would otherwise have filled.

## Testing

**Two servers with disjoint synthetic libraries** (ports 3366/3377, both 6.25.0):

```
[dj] multi-server: 2/2 answered, best 0.9817 from xs-one
```

Then with xs-one blocked by its own minimum-rating filter — proving per-server filters apply per server, and that a cross-server pick is enqueued correctly:

```
[dj] multi-server: 1/2 answered, best 0.7974 from xs-two
queue [1] Alpha Track  server=xs-two  url=http://127.0.0.1:3377/media/testlib2/...
```

The queued item carries the *winning* server's name and port — the stream URL, album art and queue extras are all built from it rather than the DJ's own server.

**Panel verified on an emulator**: the switch at the top, "The session" holding the shared settings, "Each library" holding the moved dropdown (listing both servers, current one highlighted), the participants readout, and minimum rating + genre beneath it.

`flutter analyze` at the pre-existing 16-issue baseline; `flutter test` 317/317.

> [!NOTE]
> The 6.25.0 floor carries the same caveat as the track-length window: the server params are on mStream master but in no tag yet, so that number is the expected next minor rather than an observed floor.

**Not exercised:** the model-mismatch exclusion path. The server-side 400 is verified in #929, but faking a second embedding model locally wasn't practical — the app's handling of it is reasoned, not run.

---

### Update — the two fixes the webapp's smoke found (mStream #946)

The webapp half of this feature was smoked on a live two-server rig and exposed two defects this branch shared. Both are fixed in the latest commit.

- **Foreign paths in the single-server body.** A cross-server session's history holds paths from every server that contributed a pick, and the single-server body sent the whole ring as `similarTo` to the DJ's own server. A filepath only names a row on the server holding it, and every discovery route answers a uniform `404 Track not found` for one it cannot resolve, so the first fallback pick after a peer's win stopped the session (and the sonic degradation then suppressed sonic for the rest of it). The body now carries only paths the DJ's server owns (`historyOwnedBy`, pure and unit-tested); when the whole recent history lives elsewhere it falls back the way a cold start does: the explicit seed, else the playing track. The cross-server seed still averages every anchor, and the multi-server path is gated on those anchors rather than the local ones, so a session that has wandered onto peers keeps running cross-server.
- **All-repeat rounds.** A server answers with a random row from its pool, so a round where every answer is a repeat or blocked is now asked again, up to three times, with each server's returned `ignoreList` fed back (transient, only the winner's cooldown advances), the way the single-server loop retries, instead of dropping straight to that loop.

`dart analyze` is clean on the changed files; the pure suites are 50/50 (2 new). A device run against two servers is still the missing validation.

Server side: #929 and #946 are merged; the next mStream tag must be **6.26.0** for this branch's version floor to line up.


IrosTheBeggar:
Reviewed alongside the server half (mStream #929, now rebased with route-level tests) and pushed one commit on top. The design holds up — vector seed averaged across owners, best cosine wins, model mismatch drops the server for the session, everything falls back to the single-server pick — but three things needed fixing:

- **`ignoreList` was shared across servers.** It carries server-issued track *ids*, so the one list (server A's ids, replaced by whichever server won) meant nothing on B, could exclude unrelated tracks there, and the cooldown never followed the DJ. Now per-server, with the DJ's own server still sharing the single-server path's list so a fallback pick continues the same cooldown. Only the winner's list advances.
- **`ignoreList` can't keep a vector seed off the playing song by itself.** On the server it is a soft cooldown that yields to the full pool once it covers every candidate, and a vector carries no hashes to exclude — unlike the filepath form, whose seeds are dropped from the pool server-side. The pick now refuses an answer that is the playing track or a recent anchor (keyed server + path) and takes the next-best server. A server-side hard exclusion is on the table over on #929 if you'd rather have both.
- **The 6.25.0 floor was wrong.** 6.25.0 shipped on 2026-08-30 without #929, so the vector seed lands in the next release; floor moved to 6.26.0. Before, a 6.25.0 server would have been offered a session it silently sat out of, and the participants readout would have counted it.

Also moved the centroid math into `lib/util/seed_vector.dart` (which `fetchEmbeddings` now decodes through as well) with unit tests — averaging, scaling, cancelling anchors, wrong lengths, the wire round trip, the little-endian assumption. `flutter analyze` clean for the touched files; 48 unit tests green.

Not done in this pass: a device run. The original commit was checked on an emulator and against two servers; the changes here are confined to the pick's bookkeeping, but the two-server check is worth repeating before merge. Still blocked on #929 merging and shipping as 6.26.0 for the floor to be real.

IrosTheBeggar:
### Rebased onto master (2026-09-06)

Previous head `54e07a15` → the three original commits rebased + one fix commit, `c0eeed2`.

**Conflicts** were confined to `lib/media/audio_stuff.dart` and one import in `api.dart`: master's `_resetAutoDJSession()` (#130) replaced the inline fresh-session block — the multi-server state (seed owners, excluded set, per-server ignore lists) is now cleared there — and `_queueAutoDJSong` became async, so the cross-server pick awaits it.

**What master changed under this branch:** federated peers are now virtual `Server`s in the list (#129/#141/#142) with direct transport (#150–#152). The fix commit makes the mode federation-aware rather than merely compiling:

- `Server.authToken` on the wire (the parent's JWT on the proxy path, the guest token when direct) instead of `jwt`, which is null for a peer.
- A federated peer takes part like any other server — mStream #946 put `random-songs` and `discovery/local/embeddings` on the federation allowlist, and its pick streams the way a browsed peer track does. Hidden / missing peers don't. `AudioPlayerHandler.canJoinMultiServer` is pure and unit-tested.
- Panel: selectable-server counts, `displayName` for a peer, no min-rating tile for a peer (a key has no stars — the server skips `minRating` for it), and a warning under the switch when sonic similarity is off (the mode does nothing without it; new l10n string in every locale).
- The fan-out honours the DJ lane epoch (a clear or server switch mid-pick drops the answer *and* the single-server fallback), learns Joi rejections the way the single-server loop does (a server rejecting `similarToVector` is dropped for the session instead of 400ing every pick), and a landed pick clears the owed-pick flags.

So the app's "Play from every server" is a superset of the webapp's "Play from federated servers": the user's own servers *and* any federated peers, in one session.

**Verified:** `flutter analyze` at the 17-issue baseline (same set as master); `flutter test` 534/534 (7 new).

**Still not run:** a two-server emulator/device session on the rebased code. The server has a `test-fake` 4-d discovery model now (`scanOptions.discoveryModel: 'test-fake'`), so the federation rig can do it: parent + peer, sonic on, this switch on → expect `[dj] multi-server: 2/2 answered`. The 6.26.0 floor is still an assumption — mStream master's `package.json` reads 6.25.0 today and no tag carries #929.

**Known limits worth a follow-up:** nothing dials a tunnel for the fan-out, so a Quick Connect server or a direct peer only answers while its tunnel is already up; and a server that owns no anchor is screened for its embedding model by the 400 rather than up front (a federated peer could be screened via `federation/health.discovery.modelId` through the proxy, as the webapp does).

IrosTheBeggar:
### The two follow-ups, done (`f60ab8a`)

**Tunnels follow the DJ.** `ServerManager` gained a third source of tunnel targets, `djFanOutServers`, read live from the audio handler: while a multi-server session is armed, every server it may ask keeps a tunnel — a Quick Connect server, or a direct peer (plus its parent while the peer is not direct yet), exactly as the queue does for its tracks. Released when the DJ or the mode is switched off, or a server is dropped from the session. A server whose discovery flag was never learned is targeted too, since reaching it is what fills the flag in. A pick waits up to 8 s for a tunnel that is still dialing (only the first pick after the mode comes on usually pays this); one that is down with no dial in flight is left to its retry timer and simply does not answer that pick.

**Model handshake up front.** The fan-out now reads `GET /api/v1/federation/health` → `discovery.modelId` per server before sending a vector, as the webapp does and as #929 asks of a caller. The route answers a logged-in user on the server's own wall, a peer through the parent's proxy, and a guest over a direct tunnel alike. Answers are cached five minutes (one minute for "no model", so a finishing scan is noticed soon); the anchors' owners are screened for free from their embeddings answer. A server indexed with another model is dropped for the session and the readout counts it out (string updated in all locales); one with no model right now sits the pick out. The 400 stays as a backstop for a stale cache. When only the DJ's own server is left, the single-server pick runs instead — it seeds by filepath, which the server can exclude from its pool.

**Verified:** `flutter analyze` at the 17-issue baseline; `flutter test` 540/540 (6 new: the pure model cache and the unknown-discovery target rule).

**Still not run:** the two-server emulator/device session; that now also covers "second server is Quick Connect, mode switched on → its tunnel dials and the next pick answers 2/2".

IrosTheBeggar:
### Smoke round on real devices (2026-09-06, head `26c99f5`)

Two local mStream servers from the server's master (`77f7b57a`, reporting 6.26.0), both scanned under the server's `test-fake` discovery model (8-d, deterministic). New script: `smoke/android/autodj-multi-server.sh` — federated peer, then own servers, then a Quick Connect server that is only a fan-out candidate.

| Target | Run | Result |
|---|---|---|
| Galaxy S25 (Android 16) | `autodj-multi-server.sh` — federated / own servers / Quick Connect fan-out | **30/30**: model handshake logged, `multi-server: 2/2 answered` in every phase, the winner's track queued with `djPick`+`djSonic` and playing; B's tunnel dialed 3.2 s after launch from the armed session alone |
| iPhone 17 Pro simulator (iOS 26.5) | federated phase by hand | peer's own direct tunnel dialed at launch (`dj-restore`); three picks, two won by the peer (`best 0.9796 from peer-rig-peer-a`) and streamed from `127.0.0.1:<peer port>` |
| iPhone X (iOS 16.7.16, debug over USB) | Quick Connect fan-out at launch | `[autodj] restored on rig-a` → `tunnel up … path=direct in 3190ms (dj-restore) for=iroh-rig-b` |
| Galaxy S25 | `run-android.sh` (10-min soak) | session-actions 2/2, launch-matrix 7/7, switch-during-outage 3/3, media-resumption 3/3, dead-zone 7/7, playback-soak 2/2; bt-late-pause skipped (no headset connected) |
| iPhone 17 Pro simulator | `ios/sim-launch.sh` | 3/3 (with a Quick Connect default; the resume check needs one) |
| "smoke" AVD (Pixel 7 profile) | `federation-rig.sh` HTTP | 15/16 — the one miss was the emulator's back-tap coordinate (the peer answered `db/album-songs → 200` over its own tunnel with the parent dead); the check now accepts either peer route |
| "smoke" AVD | `federation-rig.sh` Quick Connect | **21/21** — proxy path, guest ticket, the peer's own tunnel (`mode=guest`), URLs moved, the parent's tunnel released, parent killed mid-track (next track + peer API 200), renewal in place with the parent re-dialed for the access call, tunnel kept across a server switch, relaunch onto the peer's fresh port, key revoked → parked with no hot loop |

**Two bugs the smoke found, fixed in `1bad572`:**
- A federated peer never qualified for the fan-out: the app pins a peer's discovery flag to false (`_applyFederatedDefaults`, because the similar-tracks and sonic-path routes are off the federation allowlist) and the candidate rule read that flag. The two routes the fan-out uses *are* allowlisted, so a peer now qualifies on version and selectability and `federation/health` decides whether it has a model to answer with.
- The launch-time DJ restore (`restoreAutoDJ`, deliberately not the `setAutoDJ` action) never reconciled the fan-out tunnel targets, so nothing dialed at launch. It nudges now — the `(dj-restore)` reason above is that path.

Also new: the fan-out logs why it is not running (once per reason per session) and each server's health answer once per cache life, so a drive log can explain a session that stays single-server.

**Not covered:** the federation rig on the Galaxy itself — after the soak the phone was locked (screen off long enough for the keyguard) and no adb unlock took, so both rig modes ran on the emulator instead; the model-mismatch path (needs a second fake model server-side); the DJ-off tunnel release (needs the panel toggle, which the script does not drive).

**Rig notes:** the scanner does not follow a directory of symlinks (peer A got 0 tracks on the first attempt — real folders now); the checkout must *report* ≥ 6.26.0 (the rig's worktree has `package.json` bumped for the run); `sonicMinSimilarity` 0.05 for the fake model.

IrosTheBeggar:
**Addendum — federation rig on the Galaxy S25 itself** (phone unlocked by hand; head `26c99f5`):

| Run | Result |
|---|---|
| `federation-rig.sh` HTTP | **16/16** — proxy path, guest ticket, the peer's own tunnel (`mode=guest`, 3.2 s), 24 upcoming URLs moved to the peer's loopback, parent killed mid-track (next track + peer API 200), renewal in place, key revoked → parked, no hot loop |
| `federation-rig.sh` Quick Connect | **21/21** — the above over the parent's Quick Connect tunnel, plus: parent's tunnel released once the peer went direct, re-dialed for the access call at renewal, the peer's tunnel kept across a switch to a standard server, restored queue plays after a relaunch, revocation parks cleanly |

One environment note for anyone re-running: the S25 locks 60 s after its screen times out and `wm dismiss-keyguard` does not clear that keyguard, so a script whose first tap comes late (the Quick Connect rig's reconcile allowance) needs the screen kept on — the smoke helper now sets `svc power stayon usb` for the run and reverts it.

IrosTheBeggar:
**DJ-off release, verified on the Galaxy S25** (`389caae`): with the session armed on A and B a Quick Connect candidate whose tunnel came up from the arm alone, switching the DJ off from the mini-player pill logged `[iroh] tunnel stopped (no-target/dj-off) port=… for=iroh-rig-b` within seconds, and B was not re-dialed in the 35 s after. Folded into `smoke/android/autodj-multi-server.sh` as phase 4 (skipped when the phase-3 pick came from B, whose queued track keeps its tunnel by design). Final run of the four-phase script: **32/32**.

Merging with the history intact: three original commits, the rebase fix, the two follow-up fixes, the smoke round with its two bug fixes, and the harness fixes.

PR #129, opened 2026-08-30
feat(federation): federated servers — model, transport, and the server picker

IrosTheBeggar:
Phases 1, 1b and 2 of the plan this PR also adds, [`FEDERATION_PLAN.md`](FEDERATION_PLAN.md). Federated servers are now addressable, browsable and selectable; Phases 3–4 (capability gates, the iroh transport split) follow separately.

Server side is merged: the stream proxy shipped in 6.24.0, and mStream #927 (browse) / #932 + #934 (layered `GET /api`) are on master. Verified each contract against merged master rather than the PR heads.

## The shape

A federated server is one **peer** of another configured server, reached through that parent's browse proxy instead of by a URL of its own. It's an ordinary entry in `serverList` — which is what makes the picker, queue restore and the `media/<localname>` download folder work without a parallel code path — but it owns no transport, credentials or peer list.

The app's request layer already funnelled through three functions, so teaching those three about federation is the whole of it:

| | rewrite |
|---|---|
| `Server.apiUri` | `parent.apiUri('<prefix>/api<location>')` |
| `buildServerStreamUrl` | `<prefix>/stream/…` |
| `buildAlbumArtUrl` | `<prefix>/art/<one segment>` |

Delegating `effectiveBaseUrl` to the parent means **an iroh parent works with no federation-specific code**: the request rides the parent's loopback tunnel and `__lt` is applied once, to the outer URL. The proxy forwards no query params, so it can't ride through to the peer.

Credentials needed their own accessor. A federated server has no JWT, so `Server.authToken` returns the parent's and the 27 consumption sites move onto it; `jwt` stays the stored field.

## Phase 1b: capability refresh moves to `GET /api`

`getServerPaths` now issues one `GET /api` for version, server-wide `features` and the caller-scoped `user` block. That's also the **only** capability route a peer answers — ping 403s a federation key — so the federated and normal paths are now the same code rather than siblings.

Pre-#932 servers answer the same route with the old flat shape and fall back to ping. They're detected by the absence of `user`, **not** the absence of `features` — the old response carried a `features` key too (`{subsonic}`). Playlists left that payload deliberately in #932 (a resource, not a capability) and come from `/api/v1/playlist/getall`.

## The trap this design has to avoid

`_applyFederatedDefaults` pins transcode and the four discovery flags false *after* parsing whatever the peer said about itself. A peer's answer describes what **it** can do; the federation allowlist decides what we can reach through the proxy, and it's the narrower of the two. `/transcode` and every `/api/v1/discovery/*` route are off it, so a peer reporting working ffmpeg would otherwise light up UI whose requests can only 403.

## Peer lifecycle

`refreshFederatedPeers` mirrors the parent's list rather than owning it — peers are the parent admin's data. New ones appear, renames follow, and a peer that stops being listed is **flagged, not deleted**: a queued track and a downloaded file both point at its localname, so deleting the record would strand them. Removing a parent cascades to its peers; renaming one re-points them first.

Peer records persist a minimal stub because `QueueStore` resolves tracks by localname at launch, before any network call. A peer's localname is its **own** (`peer-<slug>`), never derived from the parent's — the parent's is user-editable, and deriving would force a download-folder migration for every child on a rename.

## Phase 2: the picker and the nav screen

Peers already appeared in the picker once the reconcile added them, since the menu maps over `serverList`. What they needed was to *read* as peers rather than as servers with a broken URL — indented under the server they're reached through, with a hub icon and a "via &lt;parent&gt;" line. Rows render `Server.displayName` (a peer's reported name; the URL for everything else), and the app-bar subtitle follows.

A peer whose parent has stopped listing it is skipped in the menu and excluded from the "&gt;1 server" visibility test. The menu value stays the index into the **full** list, which is what `changeCurrentServer` takes.

On the nav screen, **Playlists** and **Rated** are dropped for a peer — every `/api/v1/playlist/*` route and `db/rated` are off the allowlist, so they could only 403 — and a leading note row says why. The note is an ordinary `DisplayItem` of a new `'note'` type: `handleTap` has no branch for it and no fallback, so it's inert, and with a subtext it lands on the existing 74px `_rowExtent` without touching the list's extent math.

## Fixed in passing

All three surfaced by making these paths federation-aware:

- **`browser_toolbar` hand-assembled a `/media` download URL** instead of using the builder. On a *public iroh* server that produced `?&__lt=…`, since the loopback suffix assumes a token precedes it. Replaced with a new `buildServerDownloadUrl` (original bytes, never transcoded), pinned by a test.
- **Federated servers bypass `addServer`**, so nothing created their download folder. Extracted `_ensureDownloadDir` and called it from both doors.
- **`removeServer`'s active-server fallback keyed on identity** with the removed server, so an active server swept out as a peer of its parent left `currentServer` dangling. Now keys on list membership.

## What's left, and one finding

Phase 4 turned out bigger than the plan first had it, and the PR updates the plan to say so. A federated server has `connectionType == 'http'`, so **`isIroh` is false for it even when its parent is iroh** — and roughly 12 of the 29 `isIroh` call sites are really asking "do this server's bytes travel over a loopback tunnel?" (`cast_origin`, five in `audio_stuff`, `downloads`' stale-port rotation, art rebinding in `queue_store`/`playlist`, `makeServerCall`'s retry). Peers of an iroh parent break at every one.

It's now scoped as a `transportServer` / `isIrohTransport` split with a per-site audit, and marked **blocking release for iroh users** rather than cleanup. Blast radius is bounded — with a plain HTTP parent every one of those sites answers correctly today — but the failure mode is nasty, since browsing works while playback and casting break in ways that don't point at federation.

## Testing

23 new tests: `test/objects/federated_server_test.dart` (identity, JSON round-trip including a pre-federation server file, credential delegation, the three URL rewrites, iroh-parent composition, the `?&__lt=` regression) and `test/singletons/federated_nav_test.dart` (section gating both ways, the inert note row). Full suite **340 green**; analyzer at its 17-issue pre-existing baseline.

Not yet exercised against a live two-server rig — the peer reconcile and the `/api`-through-proxy call want an on-device pass.

### DIFF
--- a/FEDERATION_PLAN.md
+++ b/FEDERATION_PLAN.md
@@ -1,0 +1,245 @@
+# Federated server support
+
+Bringing mStream's federation feature to the mobile app: browse and play a
+paired peer's library from inside the app, using the existing multi-server
+machinery rather than a parallel one.
+
+Server side is already done and merged:
+
+| PR | What it added |
+|---|---|
+| (pre-existing, 6.24.0) | `GET /api/v1/federation/peers/:id/stream/*path` — peer audio, ranges forwarded |
+| [#927](https://github.com/IrosTheBeggar/mStream/pull/927) | the browse half: peers projection, API proxy, art proxy, `federationBrowse` ping flag |
+| [#932](https://github.com/IrosTheBeggar/mStream/pull/932) | layered `GET /api/`, added to the federation allowlist |
+| [#934](https://github.com/IrosTheBeggar/mStream/pull/934) | `/api/` back behind the auth wall; version returns to every response |
+
+---
+
+## 1. What the server offers
+
+### Routes reachable from the app
+
+| Route | Purpose |
+|---|---|
+| `GET /api/v1/federation/peers` | `{peers:[{id,name,lastSeen,lastStatus,useDiscovery}]}` — any logged-in user, never credentials |
+| `ALL /api/v1/federation/peers/:id/api/<local-path>` | one allowlisted read, executed on the peer |
+| `GET /api/v1/federation/peers/:id/art/<file>?compress=` | that peer's album art |
+| `GET /api/v1/federation/peers/:id/stream/<path>` | that peer's audio |
+
+Gated by the ping flag `federationBrowse` (federation enabled **and** at least
+one peer). Flags, never probes — the house rule on both sides.
+
+### Three constraints that shape everything
+
+**The path prefix is literal.** `/api/v1/federation/peers/{id}/api` + the
+normal local path, so `…/peers/3/api/api/v1/db/albums`. The doubled `api` is
+correct.
+
+**Query params are dropped.** The API proxy forwards none; art forwards only
+`compress`. Deliberate — the app's `?token=` is the *parent's* JWT and
+forwarding it would hand a peer a working credential for the parent. Every
+call the allowlist covers puts its params in a POST body or the path, so
+nothing is lost today. Any future call that relies on a query string will
+silently lose it.
+
+**The allowlist is the entire feature surface**
+([`federation-auth.js`](https://github.com/IrosTheBeggar/mStream/blob/master/src/api/federation-auth.js)).
+
+Reachable: `GET /api` · `GET /api/` · `db/status` · `db/metadata{,/batch}` ·
+`db/artists` · `db/artists-albums` · `db/albums` · `db/genres` ·
+`db/genre-songs` · `db/album-songs` · `db/recent/added` · `db/search` ·
+`file-explorer{,/recursive,/m3u}` · `GET /media/*` · `GET /album-art/*`
+
+Not reachable, so the app must hide it: **all `playlist/*`** · `db/rated` ·
+`rate-song` · **`db/random-songs` (Auto DJ is impossible)** · `/transcode` ·
+`/api/v1/lyrics` · `/api/v1/share` · `torrent/*` · `album-art/search` ·
+**`/api/v1/ping`** (403s a federation key — pinned by test).
+
+### `GET /api` is the one capability call
+
+Post-#934, a federation key gets the full response: `server` (version),
+`features.{discoveryReady, discovery, discoveryP2p, transcode,
+supportedAudioFiles}`, and a key-scoped `user.{username, admin, federation,
+vpaths, noMkdir, noUpload, noFileModify, federationDiscovery, vpathMetaData}`.
+
+Because ping 403s a federation key, this is the *only* capability route for a
+peer. Call it as `/api` without the trailing slash — both spellings are
+allowlisted, but the bare form sidesteps how Express 5's `*path` wildcard
+hands a trailing slash to the proxy's segment join.
+
+> **Trap:** a peer's advertised capability is not what the app can reach. A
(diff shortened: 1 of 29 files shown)


IrosTheBeggar:
## Rebased onto master + review pass (2026-09-02)

**Rebase.** Onto `35a6035` (after #137/#138/#139). Three conflicts, all mechanical:
- `api.dart`: master's `_direct` client + 15 s timeouts kept, with the PR's `authToken` accessor.
- `stream_url.dart`: the federated branch returns before master's iOS codec fallback (a peer cannot transcode anyway).
- `server_list.dart` `getServerPaths`: the PR's layered `GET /api/` refresh wins — the version rides on that call, so master's separate `fetchServerVersion` there was redundant. (`_warnIfBelowFloor` at add time still uses it.)

**Fix commit (`1268bfb`) — the first Phase 4 site turned out to be the manager itself.** `_tunnelTargetServer()` keyed on the browsed server's own type, so selecting a peer of an iroh parent (`isIroh` false) *released the parent's tunnel* and every peer request resolved to the unroutable origin: browsing failed, not just playback. The queue listener had the same blind spot (a queued peer track never held its parent's tunnel). Added `Server.transportServer` / `isIrohTransport` (the plan's Phase 4 primitives) and resolved through the transport at: the tunnel target, the launch wait for an iroh default, the resume server's pre-warm, `_settleTunnelForSwitch`, the tunnel-follows-queue listener, and the banner's "browsing the tunnel" test. Also `_pingPendingFor` is a set re-run per transport (one slot lost the parent's refresh when peers overwrote it).

**Verification.**
- `flutter test` 455/455 (+5 for the transport accessors), analyzer at the 17-issue baseline.
- Galaxy S25: `smoke/android/launch-matrix.sh` 7/7 on the rebased manager; iOS simulator `sim-launch` 3/3 against demo.mstream.io over Quick Connect.
- **Live two-server rig** (new `smoke/android/federation-rig.sh`, two mStream 6.25.0 instances paired over the federation endpoint): peer reconciled into the list, picker row, read-only nav screen, albums + cover art through the browse/art proxies, a track played through the stream proxy (loading → ready → PLAYING), downloads landed under `media/peer-rig-peer-a/…`. Same over an **iroh parent** (Quick Connect on the parent): peer reconciled through the tunnel, browsed and played, the parent's tunnel survived selecting the peer and a switch to a standard server while a peer track played (10/10).

**Still open before a release (not blocking the merge of the proxy path):**
- Phase 3 gates: Auto DJ / rating / share / lyrics on a peer answer 403 from the server; nothing hides them yet, and Auto DJ can pick a peer as its server.
- Phase 4 sites outside the manager (`audio_stuff` URL rebinding ×5, `downloads` stale-port rotation, `cast_origin` relay, `queue_store`/`playlist` art rebinding, `api.dart` retry, `auto_browse` tunnel check): queued peer URLs keep a stale parent port after a tunnel rebuild or relaunch, so a queue saved against an iroh parent needs a re-browse next session.
- Cosmetic: the peer's nav screen renders as a list (the note row), the parent's as the tile grid.
- Two entries for the same physical server (HTTPS + Quick Connect) each list the same peers → two peer rows with a `-2` localname suffix, by design.

PR #132, opened 2026-09-01
fix(quick-connect): sign in on a pushed page, not under the fold

IrosTheBeggar:
## Problem (user feedback)

In the Quick Connect flow, the credential step was easy to miss:

- **Pairing code + Test Connection** (private server): the sign-in fields appeared as an inline reveal at the *bottom* of the long, scrollable add-server page — under the fold on most phones. The test reported success and, visibly, nothing happened next.
- **Tapping a server under "Servers on your local network"**: credentials came up in a modal bottom sheet — cramped behind the keyboard and, per feedback, missed outright.

## Fix

Both flows now push a dedicated **full-screen sign-in page** (`IrohLoginScreen`) the moment credentials are needed — an unmissable next step.

- The page runs the login round-trip itself (through the live tunnel for the code flow; against the discovered LAN address for a tapped server), so a wrong password shows an **inline error and retries in place** instead of bouncing the user back a screen.
- On success it pops with the credentials and the callers finish exactly as before — the follow-up login inside `_signInAndSaveIroh` / `_connectDiscovered` runs against just-validated credentials and stays as a backstop, keeping every save/error path untouched.
- Backing out is safe: the tunnel stays up, and the Quick Connect tab keeps a **Sign in & save** button as the way back in.
- `_LanLoginSheet` (the bottom sheet) is deleted. No new localization strings were needed — the page reuses `lanLoginTitle` / `irohSignInHeader` / `irohSignInSave` / `irohSigningIn`.

## Verification

- `flutter analyze` clean; all 333 unit tests pass.
- End-to-end on an Android emulator against a live dev server (auth user + `iroh.enabled` + `shareCodePublic`): pasted a real pairing code → Test Connection dialed the actual tunnel → **the sign-in page auto-opened** → wrong password showed the inline HTTP 401 banner and stayed put → correct password signed in through the tunnel and saved the iroh server (JWT persisted, server switched current, app proceeded to Quick Setup).
- The LAN-tap path shares the same page and validate mechanics but couldn't be exercised on the emulator (mDNS doesn't cross its NAT) — worth one real-phone tap while smoking this.

### DIFF
--- a/lib/screens/add_server.dart
+++ b/lib/screens/add_server.dart
@@ -24,6 +24,7 @@ import '../singletons/browser_list.dart';
 import '../singletons/migration_manager.dart';
 import '../singletons/downloads.dart';
 import '../theme/velvet_theme.dart';
+import 'iroh_login_screen.dart';
 
 class AddServerScreen extends StatelessWidget {
   const AddServerScreen({super.key});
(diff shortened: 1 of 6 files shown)


IrosTheBeggar:
Follow-up from phone smoke: adding a v6.25 server toasted "older than 5.5". Root cause was two conflations — a FAILED version probe read the same as a 404 (the only answer that really means pre-5.4.2), and the just-saved iroh server was missing its `tunnelToken` until after `addServer`, so the add-time probe's shim auth was connection-pool luck. Fixed in 35bdc7f: `fetchServerVersion` now reports `{version, notFound}`, the warning routes through a pure `floorVerdictFor` (unit-tested — failed probe stays quiet), and the iroh server carries its token from construction so the version lands immediately. 337 tests pass.

PR #133, opened 2026-09-02
fix(iroh): keep Quick Connect alive across dead zones

IrosTheBeggar:
## Summary

Quick Connect (iroh) stayed down for minutes after a cellular dead zone and did not come back when service returned; with no downloaded tracks the app went silent while the notification still said "playing". Two drive logs from v0.35.0 were reconstructed line by line and the tunnel lifecycle audited; 17 findings were confirmed adversarially, two of them critical. The full write-up (timelines, root causes, fix plan, device script) is in a private report: https://claude.ai/code/artifact/1fb7da36-95d4-48f4-ba7f-b442d0be148d

**Root causes**

1. `handleNetworkChange` probed the loopback on every connectivity event / app resume **without reading the native status**. During a reconnect the shim fails every loopback request within a second, so the probe always failed, and a failing probe stopped the tunnel and started a fresh one with a single cold dial — replacing a supervisor that re-dials forever on the same port with one 33 s attempt (log: `resume probe failed` 15:08:50 → `tunnel start failed` 15:09:23, exactly 8 s relay wait + 25 s connect).
2. A failed cold dial was **terminal**: `tunnelPort` went null (every URL → `127.0.0.1:0`) and nothing retried. Only a transport-type change, an app resume, the banner's Retry, a playback error, a browse or a download re-enqueue could re-enter — none of which fire on a screen-off Android Auto drive.
3. The probe ran 0 s after the `network_change` nudge with a 6 s budget while iroh was still migrating paths, so every Wi-Fi↔cellular hand-off rebuilt a healthy tunnel (port + token rotation → audible player reload, stale art, killed downloads). Every rebuild in both logs was one of these.
4. Auto DJ never waited for the tunnel; an outage-drained queue ended in `stop()`, which set `_intentionalStop` and disarmed every later tunnel-back heal.

(The evening log's 5.5-minute silence turned out to be end-of-queue with Auto DJ off — v0.35.0 predates 46e53ba, already on master.)

## Phase 1 — Dart only (this PR)

- **`lib/singletons/tunnel_policy.dart`** (new): pure, unit-tested rules. Never probe or drop a reconnecting tunnel; probe a connected-reporting one twice (+3 s, +10 s) before acting; ≤1 hard rebuild per 60 s; retry 5/10/20/40/60 s then 5 min after a failed dial; a rejected pairing never re-dials on its own; a watchdog for a non-converging supervisor (120 s when the relay is known reachable, 5 min when the binary cannot say, never in a known dead zone).
- **`server_list.dart`**: single-flight `handleNetworkChange` around the policy (a Retry tap that lands mid-run keeps its user semantics); tunnel identity *and* status re-checked inside `_tunnelChain` before any drop; a retry timer so "no tunnel and nothing scheduled" is unreachable; a recently-failed gate so other callers (DJ picks re-fired every ~30 s, browse, downloads, playback errors) cannot queue cold dials past the backoff; launch dial bounded to 12 s with the skipped ping re-run on tunnel-up; lifecycle logging (`tunnel up port=… path=… in …ms (reason)`, every stop with reason, status edges, probe outcome + exception class).
- **`main.dart`**: every connectivity result logged, 1.5 s debounce, `[none]` cancels a pending run, Retry on the reconnecting banner too.
- **`audio_stuff.dart`**: an Auto DJ pick waits for the tunnel and is remembered when it can't; a queue that ends mid-outage **parks** (paused, play intent kept) instead of stopping, retries the pick immediately when the tunnel is in fact serving, and resumes from the park when the pick lands (resume is a property of the park, and the append is awaited before the seek); the park is released by any play/pause/skip/seek so it can never stop something the user started later; album-art URLs are rebound after a rebuild (downloaded items included); automatic resumes honour a 90 s interruption window; player state transitions and ≥3 s rebuffers are logged.
- **`cast_origin.dart`**: `rebindLoopbackArt` (pure, tested).
- **`iroh_tunnel.dart`**: optional lookups for the Phase 2 native symbols (`force_reconnect` / `drain_events` / `relay_online`), so the committed `.so` / xcframework keep working and Phase 2 slots in without Dart changes.

The iOS thaw case from 98d06c8 (native says connected on a dead QUIC connection) is preserved: both probes fail → hard rebuild, now backed by the retry loop.

**Deliberately unchanged:** an iroh playback error inside the ≤30 s window where the native status still says connected still takes the plain-HTTP retry walk (~35–45 s) before a ground-truth probe runs. Phase 2 (Rust: in-place reconnect on the same port, event-driven backoff, events ring, relay state) is designed and not started.

## Verification

- `flutter analyze`: clean apart from 17 pre-existing hints in unrelated files.
- `flutter test`: 404 pass, 42 new (`tunnel_policy_test`, `queue_end_action_test`, `rebind_loopback_art_test`).
- A three-lens review of the diff (lifecycle / playback / scenario replay) found 18 issues, all fixed before the first commit.
- **Smoke-tested on four targets, 17 scenarios, every recovery automatic** (details in the PR comments):
  - Android emulator vs demo.mstream.io: dead zone mid-track, Retry tap in airplane mode, unbuffered-track skip during an outage, queue end during an outage with Auto DJ, 10 s blip, launch mid-outage, steady state.
  - iOS simulator: launch / Auto DJ, 5-minute background then resume, wrong-secret pairing code.
  - Galaxy S25 over real cellular against the author's server: Wi-Fi→cellular hand-off, 70 s airplane mode, Retry tap, queue end inside the connected-but-dead window, Play after an outage-drained queue, cellular→Wi-Fi return.
  - iPhone X (iOS 15): suspension (iOS kills the loopback listener; both probes catch it and the tunnel is rebuilt), 4-minute Wi-Fi outage.
- Four defects found only by those rounds are fixed in dc858d6 and 1e065c9.

Not reproducible off a drive: a same-network signal fade with no transport event. The logging added here makes the next one readable line by line.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -164,6 +164,8 @@ class _MStreamAppState extends State<MStreamApp> with WidgetsBindingObserver {
   final GlobalKey<ScaffoldState> _outerScaffoldKey = GlobalKey<ScaffoldState>();
   StreamSubscription<String>? _castErrorSub;
   StreamSubscription<List<ConnectivityResult>>? _connectivitySub;
+  // Coalesces the burst connectivity_plus emits for one transport switch.
+  Timer? _connectivityDebounce;
   StreamSubscription<List<Server>>? _serverListSub;
   // Guards the first-run setup push: serverListStream re-emits on every server
   // edit, and the flow marks onboarding complete only once it's dismissed.
@@ -222,13 +224,28 @@ class _MStreamAppState extends State<MStreamApp> with WidgetsBindingObserver {
     // An iroh tunnel is a long-lived QUIC connection; unlike fresh per-request
     // HTTP sockets it needs an explicit kick when the device network changes.
     // Nudge it on every connectivity transition that has a usable network.
+    // Every result is logged, including [none]: the drive logs could not say
+    // what fired each network change, and a [none]→[mobile] edge is a modem
+    // re-attach the tunnel policy treats differently from a hand-off.
     _connectivitySub = Connectivity().onConnectivityChanged.listen((results) {
-      if (results.any((r) => r != ConnectivityResult.none)) {
-        unawaited(ServerManager().handleNetworkChange());
+      final names = results.map((r) => r.name).join('+');
+      appLog('[net] connectivity → [$names]');
+      if (results.every((r) => r == ConnectivityResult.none)) {
+        // A usable result seconds ago must not fire into an offline device.
+        _connectivityDebounce?.cancel();
+        _connectivityDebounce = null;
+        ServerManager().noteConnectivityLost();
+        return;
+      }
+      _connectivityDebounce?.cancel();
+      _connectivityDebounce = Timer(const Duration(milliseconds: 1500), () {
+        _connectivityDebounce = null;
+        unawaited(ServerManager()
+            .handleNetworkChange(reason: 'connectivity:$names'));
         // Resume on-device playback if a network outage had paused it (no-op
         // unless the audio handler is in the network-stalled state).
         MediaManager().audioHandler.onNetworkRegained();
-      }
+      });
     });
   }
 
@@ -510,7 +527,7 @@ class _MStreamAppState extends State<MStreamApp> with WidgetsBindingObserver {
     // resume, nudge iroh + rebuild it if it's hard-down. Fire-and-forget — the
     // native start can block for tens of seconds.
     if (state == AppLifecycleState.resumed) {
-      unawaited(ServerManager().handleNetworkChange());
+      unawaited(ServerManager().handleNetworkChange(reason: 'resume'));
     }
     // Flush the queue/position to disk when leaving the foreground, so a
     // backgrounded app that's later killed by the OS still reopens in place.
@@ -526,6 +543,7 @@ class _MStreamAppState extends State<MStreamApp> with WidgetsBindingObserver {
     WidgetsBinding.instance.removeObserver(this);
     _castErrorSub?.cancel();
     _connectivitySub?.cancel();
+    _connectivityDebounce?.cancel();
     _serverListSub?.cancel();
     DownloadManager().dispose();
     super.dispose();
@@ -579,8 +597,9 @@ class _MStreamAppState extends State<MStreamApp> with WidgetsBindingObserver {
             return _bannerStrip(
                 Icons.cloud_off, VelvetColors.warning, l.irohBannerDisconnected,
                 action: TextButton(
-                  onPressed: () =>
-                      unawaited(ServerManager().handleNetworkChange()),
+                  onPressed: () => unawaited(ServerManager()
+                      .handleNetworkChange(
+                          reason: 'retry-tap', userInitiated: true)),
                   child: Text(l.irohRetry,
                       style: TextStyle(color: VelvetColors.primary)),
                 ));
@@ -593,12 +612,17 @@ class _MStreamAppState extends State<MStreamApp> with WidgetsBindingObserver {
                   child: CircularProgressIndicator(strokeWidth: 2),
                 ));
           default: // reconnecting
+            // Retry here forces a fresh endpoint (the policy's `escalate`):
+            // the supervisor keeps re-dialing on its own, but a user who has
+            // watched the spinner long enough gets to cut it short.
             return _bannerStrip(
                 Icons.sync_problem, VelvetColors.warning, l.irohBannerReconnecting,
-                action: const SizedBox(
-                  width: 16,
-                  height: 16,
-                  child: CircularProgressIndicator(strokeWidth: 2),
+                action: TextButton(
+                  onPressed: () => unawaited(ServerManager()
+                      .handleNetworkChange(
+                          reason: 'retry-tap', userInitiated: true)),
+                  child: Text(l.irohRetry,
+                      style: TextStyle(color: VelvetColors.primary)),
                 ));
         }
       },
--- a/lib/media/audio_stuff.dart
+++ b/lib/media/audio_stuff.dart
@@ -3398,7 +3398,16 @@ class AudioPlayerHandler extends BaseAudioHandler
     // add followed by seek(index + 1) reached ExoPlayer with an index past the
     // end (IllegalSeekPositionException) and left the darwin player's index
     // out of range.
-    await addQueueItem(item);
+    // Tolerant of the platform insert throwing (Android's known just_audio
+    // IllegalArgumentException on concatenatingInsertAll): the item is already
+    // in the queue subject by then, and an unawaited add used to let the pick
+    // proceed regardless — awaiting it must not turn that into an aborted
+    // pick with no resume.
+    try {
+      await addQueueItem(item);
+    } catch (e) {
+      appLog('[dj] queue insert threw (continuing): $e');
+    }
     final resume = wasParked && _parkedForDJ;
     _releasePark();
 
--- a/lib/media/audio_stuff.dart
+++ b/lib/media/audio_stuff.dart
@@ -4,7 +4,7 @@ import 'dart:io' show File;
 
 import 'package:audio_service/audio_service.dart';
 import 'package:audio_session/audio_session.dart'
-    show AudioSession, AudioSessionConfiguration;
+    show AudioSession, AudioSessionConfiguration, AudioInterruptionType;
 import 'package:just_audio/just_audio.dart';
 import 'package:mstream_music/main.dart';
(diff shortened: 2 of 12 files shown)


IrosTheBeggar:
## Smoke-test round (2026-09-01, commit dc858d6)

Debug builds of this branch against **demo.mstream.io over Quick Connect**. Android: the headless `smoke` AVD (android-36 arm64), outages via `cmd connectivity airplane-mode`. iOS: iPhone 17 Pro simulator (no network toggle there, so steady-state / resume / rejected-code only). The physical Galaxy was not reachable over adb and the iPhone showed offline, so **the real-device round is still open**.

| # | Scenario | Result | What the log showed |
|---|---|---|---|
| 1 | Steady state, Auto DJ over the tunnel (both platforms) | ✅ | `tunnel up port=… path=direct in 1.3–3.2 s (launch)`, DJ picks append, resume-time probe passes with no rebuild |
| 2 | **65 s dead zone mid-track** (Android) | ✅ | `status connected → reconnecting` at +33 s; on service return `network change … native=reconnecting` → `leaving the reconnect to the supervisor` (×2); `status reconnecting → connected after 42 s` **same port**, playback never interrupted. No `tunnel stopped` / `start failed` / `URLs rebuilt`. |
| 3 | **Retry tap while in airplane mode** (the log-A terminal state) | ✅ | `tunnel stopped (rebuild/escalate/retry-tap)` → `start failed after 33041 ms` → `retry #1 in 5s` → a DJ-driven ensure `start skipped … retry #1 owns the next attempt` → `retry #2 in 10s` → **`tunnel up … (retry#2)` 5 s after airplane off, no taps** → `stream URLs rebuilt` → track resumed → deferred pick landed |
| 4 | Skip to an unbuffered track during the outage | ✅ | Source error → `iroh recovery … parked until it reconnects`; 3 s after service: `iroh tunnel back — resuming parked playback` + deferred pick, same port |
| 5 | **Queue ends during the outage, Auto DJ on** | ✅ | `pick deferred` → `completed → park` (no `stop`) → 6 s after service: `tunnel back — retrying the deferred pick and resuming` → `queue add` → `pick landed — resuming the parked queue` → next pick appended |
| 6 | 10 s blip | ✅ | No status edge, no rebuild; the `[none]→[mobile]` edge within 15 s is left alone, the later `[wifi]` probe passes |
| 7 | **Launch mid-outage** | ✅ | Restore parks at the saved spot; `start failed (launch)` → retry 5 s → retry 10 s (re-armed by the connectivity events) → `tunnel up (retry#2)` → `stream URLs rebuilt` → `resuming parked playback` → media Play works |
| 8 | Background 5+ min, resume (iOS sim) | ✅ | `network change (resume) native=connected` → `probe #1 passed` → nothing torn down |
| 9 | Wrong-secret pairing code (iOS sim) | ✅ after fix | `start failed … handshake rejected` → `status … → rejected` (Repair banner); **found:** the queue-restore ensure bypassed the rejected gate and forced one extra dial — fixed in dc858d6. No further dials over the following minute. |

Two defects found and fixed in dc858d6: the awaited DJ append let Android's pre-existing just_audio `IllegalArgumentException` abort a pick (now tolerated), and `setQueueIrohServer` bypassed the backoff/rejected gates. Full suite still green (403).

Still to do before merge: the same scenarios 2, 3, 5 and 7 on the Galaxy over real cellular (wireless debugging needs to be on, or a cable), plus the Wi-Fi→cellular hand-off, which no emulator can reproduce.

IrosTheBeggar:
## Real-device round — Galaxy S25 (Android 16), commit 1e065c9

The author's own server over Quick Connect (relay path), phone on home Wi-Fi then cellular, outages via airplane mode over USB adb. Keep-queue-offline was on (Wi-Fi-only), so picks made on Wi-Fi play from disk and picks made on cellular stream through the tunnel — the drive configuration.

| # | Scenario | Result | What the log showed |
|---|---|---|---|
| 1 | **Wi-Fi → cellular hand-off** while playing (leaving the house) | ✅ | `[mobile]` → `network change … native=connected` → `probe #1 passed (http 401, 3483ms)`, tunnel kept its port, nothing rebuilt, playback untouched |
| 2 | **Airplane mode 70 s on cellular**, NEXT onto a downloaded track during it | ✅ | `status connected → reconnecting` at +35 s; the skipped track fell back to its local copy; connectivity event → `leaving the reconnect to the supervisor`; **in-place reconnect on the same port** |
| 3 | **Retry tap in airplane mode** (the afternoon log's terminal state) | ✅ | `tunnel stopped (rebuild/escalate/retry-tap)` → `start failed after 33055ms` → retry 5 s / 10 s / 20 s → `tunnel up … (retry#3)` 47 s after cellular returned → `stream URLs rebuilt` → `iroh tunnel back — resuming parked playback` |
| 4 | **Queue ends inside the connected-but-dead window**, Auto DJ on | ✅ after fix | `completed → park` → the in-flight POST timed out → `tunnel says connected but the probe failed (timeout, 6008ms)` → `tunnel stopped (rebuild/dead/reverify)` → cold dial fails → retry 5 s / 10 s → Wi-Fi back → `tunnel up … (retry#2)` → `album-art URLs rebound` → `retrying the deferred pick and resuming` → `pick landed — resuming the parked queue`, **16 s after service returned** |
| 5 | Press Play after an outage-drained queue | ✅ | re-seed at the end of the last track → `completed → park` → in-flight pick landed → resumed |
| 6 | Cellular → Wi-Fi return | ✅ | retry re-armed on the connectivity event, connected on Wi-Fi in 3.3 s, deferred pick landed, downloads resumed |

**Fixed in 1e065c9 (found only on the real phone):** on this device the native status keeps saying `connected` for ~70 s after the link dies (vs ~33 s on the emulator), so a queue end in that window reached the completed handler with nothing pending and stopped. A pick is now owed from the moment its tunnel wait begins, a pick still in flight counts as owed at queue end, and a queue-end retry that fails while the tunnel claims to serve parks and runs the ground-truth probe instead of stopping.

**Numbers worth keeping for Phase 2:** the supervisor's in-place reconnect took **65 s after service returned** (108 s from the drop) — the timed-backoff cost the Rust event-driven backoff is meant to remove. A **fresh cold dial over this carrier took 20–33 s** (one failed at the 33 s budget while a Wi-Fi dial took 3.3 s), so the planned 12 s reconnect dial timeout would be too aggressive on real cellular; keep the reconnect dial at ≥25 s and get the speed-up from the relay-back wake instead.

Not covered: a same-network signal fade (no transport event) — only reproducible on a drive; the logging added here will make the next one readable.

IrosTheBeggar:
## Real-device round — iPhone X (iOS 15.6.1, Wi-Fi only), profile build, demo.mstream.io over Quick Connect

Log source: the app's own Diagnostics export (a debug build cannot be launched from the home screen on iOS 14+, and a profile build's prints don't reach syslog or `flutter attach` on this device).

| # | Scenario | Result | What the log showed |
|---|---|---|---|
| 1 | **Suspension** (pause, lock the phone ~7 min, reopen, play) — the 98d06c8 case | ✅ | On thaw the QUIC side flipped to reconnecting and the supervisor re-dialed **within 1 s** on the same port — but the app's loopback **listener had been killed by iOS during suspension**, so native read connected while every local request was refused. `probe #1 failed (refused, 5ms)` → `probe #2 failed (refused, 2ms)` → `tunnel stopped (rebuild/dead/resume)` → `tunnel up port=50190 … in 3267ms` → `stream URLs rebuilt — reloading at track 2` → playing, **14 s after reopening**, no extra taps. |
| 2 | **Wi-Fi off ~4 min while streaming, then on** | ✅ | Resume mid-outage: `probe #1 failed (reset, 5264ms, native=reconnecting)` — the status flipped during the probe and the code stood down (`leaving the reconnect to the supervisor`), no needless rebuild. Buffered track played on; the next one failed with `-1005` and parked; DJ pick deferred. Wi-Fi back → `status reconnecting → connected after 213s` **1.2 s later, same port** (Apple's network monitor is real, unlike Android's) → `resuming parked playback` + deferred pick landed → later probe passed. |

**For Phase 2 (Rust):** the iOS suspension kills the loopback `TcpListener`; today `bridge`'s accept loop exits silently on error (`Err(_) => break` in `connect_tunnel`) and the status keeps saying connected. An in-place QUIC "kick" alone would NOT fix this case — the kick must also re-bind the listener (same port), and the accept loop's exit should flip the status (or re-bind) instead of vanishing. The Dart probe's `refused` classification is what caught it here.

All three platforms now covered: Android emulator (9 scenarios), iOS simulator (3), Galaxy S25 over real cellular (6), iPhone X (2). Every recovery was automatic.

PR #134, opened 2026-09-02
feat(iroh): Phase 2 — reconnect in place, wake on relay-back, native events

IrosTheBeggar:
## Summary

Phase 1 (#133) made Quick Connect recovery reliable; this makes it fast and invisible. All changes are in the Rust tunnel crate plus its two committed binaries. The Dart side already looked the new symbols up optionally (`hasKick` / `kick` / `drainEvents` / `relayOnline`), so it lights up with no Dart changes.

**What the native side does now**

- **Reconnect in place** (`mstream_iroh_force_reconnect`): nudge iroh about the network, **re-bind the loopback listener on the same port**, then close the current QUIC connection so the supervisor re-dials immediately. Same endpoint, port and token — no URL rebuild, no player reload, no lost downloads. The app calls it after two failed liveness probes on a tunnel that still reports connected; the stop/start rebuild remains the fallback (45 s post-kick watchdog). The listener re-bind exists because iOS kills the listener during an app suspension while the QUIC connection survives (iPhone X round on #133).
- **The accept loop never exits silently again**: on an accept error it re-binds the same port in place; only a port it cannot re-bind takes the tunnel down (which the Dart side then rebuilds).
- **Event-driven backoff**: ≤10 s between attempts (was 30 s), cut short by an app kick or by the home relay coming back, reset to 1 s after a wake. On the Galaxy over cellular, the old timed backoff had let a reconnect trail service return by 65 s. The reconnect dial **keeps its 25 s timeout** on purpose — cold dials over that carrier took 20–33 s.
- **Bridges wait** up to 10 s for a swapped-in connection instead of failing in 1.2 s; a bridge that gives up drains the request and closes cleanly (FIN, not RST).
- **Native events ring** (`mstream_iroh_drain_events`): connection closed and why, each attempt with elapsed time / relay state / failure reason, listener re-binds, bridge counters — the app's 2 s status poll appends them to the diagnostics log as `[iroh-native] …`. First-dial failures now say why (e.g. `connect timed out after 25s; home relay not reached`) instead of "may be offline".
- **`mstream_iroh_relay_online`**: whether a home relay is connected, so the app's reconnect watchdog can tell a dead zone (leave the supervisor alone) from a supervisor failing with the relay up (worth a fresh endpoint).
- Dev client gains `--kick-after` / `--events`; the interop harness gains phase 7 **KICK**; `build-ios.sh` checks for 12 exported symbols; README checksums refreshed.

## Verification so far

- `cargo test`: 9 pass (new: backoff doubling/cap/wake, events ring cap + drop count).
- `interop/harness.mjs`: **ALL PASS** — JSON, Range/206, concurrency, reconnect, and the kick phase (listener re-bound in 0.3 s, `reconnected: attempt 1`, same `LOCAL_PORT` answers, server saw exactly one more connection).
- **Android emulator vs demo.mstream.io** (new `.so`):
  - Resume 16 s into an airplane-mode outage → `probe #1 failed (timeout)` → `probe #2 failed (reset)` → **`kicking the tunnel in place — port 35647 kept`** → `[iroh-native] app kick…`, `listener re-bound` → `backoff cut short: app kick` → `reconnected: attempt 2 in 8.1s` **4 s after the network returned**, same port, no rebuild, Auto DJ carried on.
  - Plain 65 s dead zone → reconnect **4 s** after service returned (Phase 1: 11 s), with the attempt/backoff narration in the app log.
  - Resume 7 s into an outage → policy stands down (recent `[none]`), supervisor heals.
- **iOS simulator** (new xcframework): tunnel up on a direct path, native events flowing, 6-minute background → resume → probe passed, nothing torn down.

## Real devices (2026-09-02) — all pass

Full table in the round comment below. Headlines, both on the **same port** with nothing rebuilt:

- **Galaxy S25 over cellular**: 75 s dead zone → reconnected 20 s after service returned (Phase 1: 65 s); queue end during the outage → parked, reconnected 8 s after Wi-Fi, deferred pick landed and playback resumed 0.2 s later; Retry tap while reconnecting → `kicking the tunnel in place … port kept`, up 7 s after Wi-Fi.
- **iPhone X**: 7-minute suspension → `probe #1 failed (refused)` (iOS reaped the listener) → kick → `listener re-bound`, `reconnected: attempt 1 in 0.1s`, play worked — no `tunnel stopped`, no `stream URLs rebuilt` (that was the Phase 1 outcome); Wi-Fi off 3 min → reconnected in the same second Wi-Fi came back. With fb2fab5 the thaw kick fires 3 s after reopening instead of 10 s.

## Follow-up from the round (fb2fab5, Dart only)

- **Retry kicks in place first.** The banner's Retry on a reconnecting tunnel still went to a hard rebuild (that branch predated the kick): two cold dials into the same dead zone and a new port 19 s after Wi-Fi — slower than not tapping. `TunnelPolicy.onUserEscalate`: kick, and only a second tap inside the 45 s post-kick window gets the fresh endpoint.
- **A "refused" first probe is definitive.** Nothing but a dead loopback listener refuses, so the 7 s wait for probe #2 only delayed every iOS thaw. `TunnelPolicy.probeIsDefinitive`.
- **Fast retry after a mid-dial network return.** A cold dial in flight when connectivity returns can still time out, and then waited the retry ladder's next rung (up to 60 s / 5 min deep into an outage). `TunnelPolicy.retryAfterFailedDial` restarts the ladder with a 2 s attempt. Unit-tested; on the device the in-flight dial recovered by itself both times the return landed early enough.

Both binaries are rebuilt from this source (checksums in `rust/iroh_tunnel/README.md`); `cargo-ndk` and the Android targets are now installed on the build Mac.


IrosTheBeggar:
## Real-device round — Phase 2 binaries (a2de0a4) + follow-up Dart fixes (fb2fab5)

Same setup as the Phase 1 round: the author's own server over Quick Connect on the Galaxy S25 (Android 16, home Wi-Fi + cellular, outages via airplane mode over USB adb, keep-queue-offline Wi-Fi-only); demo.mstream.io over Quick Connect on the iPhone X (iOS 15.6.1, Wi-Fi only, profile build, logs via Diagnostics → Share).

### Galaxy S25

| # | Scenario | Result | What the log showed |
|---|---|---|---|
| 1 | Launch on Wi-Fi | ✅ | `tunnel up port=38199 path=relay in 3775ms`, then the new native events: `[iroh-native] bound 127.0.0.1:38199, first dial OK in 0.2s` |
| 2 | Wi-Fi → cellular hand-off while playing | ✅ | `probe #1 passed (http 401, 3169ms)`, port kept |
| 3 | 75 s airplane mode on cellular | ✅ | `conn closed: timed out` at +34 s → `attempt 1 failed after 33.0s … backoff 1s` → `reconnected: attempt 2 in 32.6s` on the same port, **20 s after `[mobile]`** (Phase 1 on this phone: 65 s). Deferred DJ pick landed at once, nothing rebuilt |
| 4 | Cellular → Wi-Fi return | ✅ | probe #1 timed out mid-migration, probe #2 passed, port kept |
| 5 | **Queue end during the outage** (Auto DJ on the iroh server) | ✅ | `completed → park` → `reconnected: attempt 2 in 21.6s` 8 s after `[wifi]`, same port → `pick landed — resuming the parked queue` 0.2 s later. No rebuild anywhere |
| 6 | Retry tap while reconnecting — **a2de0a4** | ⚠️ | `tunnel stopped (rebuild/escalate/retry-tap)`: two cold dials into the dead zone, new port 19 s after `[wifi]` — slower than not tapping. The escalate branch predated the kick → fixed in fb2fab5 |
| 6b | Retry tap while reconnecting — **fb2fab5** | ✅ | `kicking the tunnel in place (retry-tap) — port 33105 kept` → `app kick: network_change, listener re-bind, close`, `listener re-bound`, `backoff cut short: app kick` → `reconnected: attempt 3 in 0.9s` **7 s after `[wifi]`**, same port |
| 7 | Launch inside the dead zone, service back mid-dial (×3) | ✅ | Twice the in-flight cold dial simply completed 2–6 s after `[wifi]`; once it had just failed and retry #1 came up 10 s after `[wifi]`. The new 2 s mid-dial retry (fb2fab5) is for a return in the dial's last seconds, which I could not hit on the device — unit-tested only |

### iPhone X

| # | Scenario | Result | What the log showed |
|---|---|---|---|
| 8 | **Suspension 7 min** (paused, locked) — a2de0a4 | ✅ | At thaw: QUIC re-dialed in 0.1 s, then `probe #1 failed (refused)`, `probe #2 failed (refused)` (iOS reaped the listener) → `kicking the tunnel in place (resume) — port 51005 kept` → `listener re-bound`, `reconnected: attempt 1 in 0.1s`; play worked. Phase 1 here: `tunnel stopped` + new port + `stream URLs rebuilt` |
| 9 | Wi-Fi off 3 min while playing | ✅ | attempts with 1/1/2/4 s backoff, `reconnected: attempt 5` **in the same second as `[wifi]`**, port 51005 kept |
| 10 | Suspension again — fb2fab5 | ✅ | thaw → `probe #1 failed (refused, 7ms)` → **`kicking the tunnel in place (resume) — port 51356 kept` 3 s after reopening** (was 10 s with the second probe) → `listener re-bound`, `reconnected: attempt 1 in 0.1s`; play worked |

Port 51005 survived the whole first iPhone session (launch → suspension → kick → Wi-Fi outage); scenario 10 ran on a fresh launch after the fb2fab5 reinstall.

### Notes

- The native status lied "connected" for 34 s on the Galaxy this round (last night ~70 s); iroh's 30 s idle timeout is the floor, so that is as good as it gets without a heartbeat.
- Residual, by design for now: a dial already in flight when service returns runs its own QUIC retransmit schedule; when the return lands in its last seconds the attempt still times out and the supervisor retries after ≤1 s. An abortable dial (cancel + fresh Initial packets on the app's network nudge) is a possible Phase 3 tweak; on this carrier a fresh dial is 20–33 s anyway, so no change now.
- Cosmetic nit in the events: a kick that lands while nothing is backing off leaves its `Notify` permit stored, so the next outage's first backoff logs `backoff cut short: app kick`. Harmless (that first backoff is skipped and the ladder restarts at 1 s); a zero-timeout drain of the permit right after `closed()` returns fixes it at the next binary refresh.
- Both phones were restored afterwards (Galaxy server order + Auto DJ target; iPhone servers.json).

PR #136, opened 2026-09-02
feat(ios): CarPlay — browse, play and queue the library from the car

IrosTheBeggar:
## Summary

CarPlay support for the approved audio entitlement. The car gets the same library tree Android Auto renders, playback goes through the audio handler like every other entry point (Auto DJ, the Quick Connect tunnel and offline copies come for free), and Now Playing is CarPlay's own template fed by what audio_service already publishes.

**One app-owned Flutter engine.** CarPlay can launch the app with no phone scene at all (phone locked, icon tapped on the car display), so the engine now starts in `AppDelegate` and the phone `SceneDelegate` attaches a `FlutterViewController` to it when the phone scene connects. It registers the engine for scene life-cycle events explicitly: with `UIApplicationSupportsMultipleScenes` on (a CarPlay requirement) Flutter no longer does that automatically (checked in the engine source). The storyboard is dropped from the scene manifest, since it would have created a second, implicit engine. An iPad "New Window" gets a placeholder and is closed again.

**CarPlay scene** (`ios/Runner/CarPlay.swift`): root list, drill-down lists with cover art fetched by `URLSession` (art URLs self-authenticate), the audio-app template stack capped at 5 (the top level is swapped instead of the push failing), Now Playing via `CPNowPlayingTemplate`, and a Queue list behind its Up Next button with tap-to-skip.

**Dart bridge** (`lib/native/carplay_bridge.dart`, channel `mstream/carplay`): serves `AutoBrowse.children` / `play` unchanged, plus the live queue and `skipToQueueItem`. On iOS the tree hands over the plain art URL instead of Android's `content://` wrapper. `play` is fire-and-forget: the handler's play future completes only when playback later pauses, and the car wants the row's completion (and its Now Playing screen) at once — found on the simulator, where Now Playing never appeared before this.

**Debug hook**: `ext.mstream.carplay` (VM-service extension, debug builds only) taps rows, goes back, opens Up Next and reports the template state, because the Simulator's CarPlay window accepts no synthetic input. Its Swift half is compiled only under `DEBUG`, which the Runner Debug configuration now defines.

**Signing**: `Runner.entitlements` with `com.apple.developer.carplay-audio` and `CODE_SIGN_ENTITLEMENTS` on the Runner target. Simulator builds need nothing else; device builds need a manually created provisioning profile with the CarPlay capability (automatic signing cannot mint one).

## Verification — iOS Simulator (Xcode 26.6), CarPlay display, demo.mstream.io over Quick Connect

| # | Scenario | Result |
|---|---|---|
| 1 | App icon on the CarPlay home screen; tap → root list (Shuffle All, Recently Added, Playlists, Albums, Artists, Files) | ✅ `[carplay] connected (dart ready: 1)`, `root: 6 rows` |
| 2 | Albums → 11 albums with cover art over the tunnel → album → 6 tracks → tap a track | ✅ `[carplay] play(…)` → queue loaded, `[play] play` → **Now Playing** with track / artist / album, transport and the Queue button |
| 3 | Up Next → Queue list (art + subtitles) → tap row 2 | ✅ `[carplay] skipToQueueItem(2)` → `track 3/6: Ghosts`, popped back |
| 4 | Shuffle All | ✅ `play(mstreamauto://shuffle…)` → queue cleared, Auto DJ picks, Now Playing |
| 5 | Recently Added / Playlists / Artists / Files | ✅ 20+ tracks / 6 playlists / 3 artists / 3 folders |
| 6 | Depth guard: Files → folder → folder → track (Now Playing at depth 5) → Up Next | ✅ `template stack at 5 — replacing the top level`, Queue shown at depth 5 |
| 7 | **Headless cold launch**: app terminated, icon tapped on the car display | ✅ `[carplay] connected (dart ready: 0)` → `dart ready (car connected: 1)` → tunnel up → root rendered → queue restored, phone still on the springboard |
| 8 | Phone UI opened on that CarPlay-launched process | ✅ full UI, mini-player showing the restored track |
| 9 | Phone scene life-cycle with the manual registration (HOME → reopen) | ✅ `[iroh] network change (resume)` → probe passed |

`flutter test`: all pass (new: `test/native/carplay_bridge_test.dart`). Analyzer clean on the touched files.

## Device testing — needs one thing from the account holder

Xcode's automatic signing cannot include the CarPlay entitlement. In the developer portal: enable **CarPlay Audio App** on the `mstream.music` App ID, create a **Development** provisioning profile for it with the test iPhones' UDIDs and your development certificate, download it, and I'll switch the Runner target to manual signing with that profile for the device build. Then test with the Toyota, or Apple's CarPlay Simulator app (Additional Tools for Xcode).

## Follow-ups (not in this PR)

- Now Playing buttons for shuffle / repeat / Auto DJ (`CPNowPlayingShuffleButton` etc.).
- Album art from servers with self-signed certificates (the Swift loader trusts only valid certs; the Dart HTTP override does not apply to it).
- `CPTabBarTemplate` (4 tabs) instead of the single root list, once the root nodes settle.
- `Main.storyboard` is now unused (dropped from the manifest) and can be deleted from the project in a cleanup.
- The iPad second-window placeholder path is untested.

### DIFF
--- a/ios/Runner.xcodeproj/project.pbxproj
+++ b/ios/Runner.xcodeproj/project.pbxproj
@@ -12,6 +12,7 @@
 		3B3967161E833CAA004F5970 /* AppFrameworkInfo.plist in Resources */ = {isa = PBXBuildFile; fileRef = 3B3967151E833CAA004F5970 /* AppFrameworkInfo.plist */; };
 		74858FAF1ED2DC5600515810 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 74858FAE1ED2DC5600515810 /* AppDelegate.swift */; };
 		7884E8682EC3CC0700C636F2 /* SceneDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7884E8672EC3CC0400C636F2 /* SceneDelegate.swift */; };
+		CA9F2E7A3B1D4C5E6F708192 /* CarPlay.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA9F2E7A3B1D4C5E6F708193 /* CarPlay.swift */; };
 		78A318202AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage in Frameworks */ = {isa = PBXBuildFile; productRef = 78A3181F2AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage */; };
 		97C146FC1CF9000F007C117D /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FA1CF9000F007C117D /* Main.storyboard */; };
 		97C146FE1CF9000F007C117D /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FD1CF9000F007C117D /* Assets.xcassets */; };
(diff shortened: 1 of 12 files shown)


IrosTheBeggar:
## Now Playing buttons + a play/pause fix (374fcd1, and the follow-up commit)

**Shuffle, repeat and Auto DJ** now sit on the CarPlay Now Playing screen. The two system buttons draw their state from the remote command center's shuffle/repeat types, which audio_service never sets on iOS, so Dart pushes the three modes (and now the playing flag) to the car on every change and the Swift side applies them. Auto DJ is an image button (SF Symbol `sparkles`, the app's own icon) with a selected state. Taps route to the same toggles the player panel and queue header use. The handler also advertises `setShuffleMode` / `setRepeatMode` as system actions now, which enables the OS remote commands underneath (and the lock-screen / Android Auto shuffle and repeat controls).

**Found while testing:** the template's main play/pause button showed "play" while the track was audibly running. On iOS the template reads `MPNowPlayingInfoCenter.playbackState` (iOS 13+, "must be set whenever the app begins or halts playback" per the SDK header), and audio_service sets it only on macOS. Fixed by setting it from the same push.

| Check (Simulator CarPlay display) | Result |
|---|---|
| Three buttons render on Now Playing | ✅ |
| Shuffle button: on → highlighted, off → plain; `[carplay] shuffle → on/off`, state follows | ✅ |
| Repeat button: off → all → one → off; the "one" badge renders | ✅ |
| Auto DJ button: off → plain, on → selected; `[carplay] Auto DJ → off / on (iroh-demo)` | ✅ |
| Play/pause glyph: pause while playing, play after pausing from the phone, pause again on resume | ✅ |

Debug hook gained `action=button&index=N`.

IrosTheBeggar:
## Siri + artist button (7e8c751)

- **Siri, "play <X> on mStream"**: `INPlayMediaIntent` handled in-app (no extension target). Resolution runs the browse tree's search (tracks → albums → artists) and hands Siri up to five candidates; the chosen node plays through `AutoBrowse.playNode`. Dry-run on the simulator (the handler minus speech): "Color Out" → album + artist candidates, played the album; "Mercurial" → both versions of the song, played; a nonsense phrase → not found. Adds the Siri entitlement, `NSSiriUsageDescription` and `INIntentsSupported`. Real Siri needs a device.
- **Artist button on Now Playing**: tapping the artist name opens that artist's albums (from the server the track came from). Simulator: "Color Out" → 5 albums, pushed at depth 5.
- Also in this PR since the first comment: Now Playing shuffle / repeat / "DJ" buttons (rendered at the car display's scale) and the play/pause glyph fix.

Device round still pending the manual provisioning profile (CarPlay Audio App + Siri capabilities on the App ID).

PR #142, opened 2026-09-03
feat(federation): Phase 3 — capability gates for federated peers

IrosTheBeggar:
Stacked on #141 (Phase 4), which is stacked on #129. Every surface whose route the federation allowlist refuses is now gated on `isFederated`, at the track level where a peer's track can sit in a queue browsed from another server.

## What changed

- **Auto DJ.** The panel's server dropdown skips peers; the queue-header toggle refuses a peer with a snack and a log line (turning the DJ *off* while a peer is browsed still works); CarPlay's toggle and the car's Shuffle All refuse with a log line; the handler's `setAutoDJ` action is the backstop for every entry point.
- **Track sheet.** No rating badge, no lyrics badge and no Add to playlist for a peer's track — Download stays. The metadata screen's lyrics chip is gated the same way.
- **Share** blocks a queue of peer tracks with its own message, ahead of the iroh "no public URL" block.
- **Torrent.** The add-torrent screen offers only the user's own servers.
- **Car root.** A peer's root has no Shuffle All and no Playlists (`AutoBrowse.rootTabs`, now test-visible).
- **Manage servers.** A peer row shows its name, a hub icon and "via <parent>" (or "No longer shared by <parent>"), and offers Info plus Hide/Show — no Edit, no pairing code, no Delete.
- **The Forget decision.** Peers are the parent admin's data, so a removed peer would come back on the next reconcile. Hiding is the durable choice: `federationHidden` is persisted and honoured by the picker and the car through `Server.isSelectable`; hiding the browsed peer moves the browser to its parent first. Forget (delete) appears only once the parent has stopped listing the peer.
- Strings: six new keys; generated locale files regenerated.

`FEDERATION_PLAN.md` Phase 3 marked done.

## Verification

| | Result |
|---|---|
| `flutter test` | 469 pass (+6: `isSelectable` and the hidden flag's round trip; a peer's car root has neither Shuffle All nor Playlists) |
| analyzer | 17-issue baseline |
| `smoke/android/federation-rig.sh`, HTTP mode | 7/7 |
| by hand on the rig (Galaxy S25, screenshots in `smoke/out/`) | Auto DJ on the peer → the snack + `[dj] refused` in the log · peer track sheet → no rating/lyrics badge, no Add to playlist, Download present · Share Playlist → "Can't share this queue … live in someone else's library" · Add torrent → defaults to the parent · Manage Servers → peer row with hub icon and "via", menu = Make Default / Info / Hide; Hide dims the row, drops it from the picker and moves the browser to the parent (`[srv] switched to rig-b`); Show restores it · **`.m3u` on the peer**: File Explorer lists `mix.m3u`, opening it POSTs `file-explorer/m3u` through the proxy (200) and lists its entries, and an entry plays — keep the rows |

Noticed, not fixed here: the m3u listing shows the `#EXTM3U` header as a row on any server (pre-existing); and the rig's *freshly enabled* Quick Connect endpoint stalled the phone's first tunnel dials in the handshake (13 s each) in several runs — worth a look on the server side, independent of federation.

### DIFF
--- a/FEDERATION_PLAN.md
+++ b/FEDERATION_PLAN.md
@@ -197,20 +197,38 @@ for a federated server, keeping File Explorer / Albums / Artists / Recent /
 Local Files, plus a read-only note mirroring the webapp's left-nav
 explanation.
 
-### Phase 3 — capability gates
+### Phase 3 — capability gates ✅ done
 
-`isFederated` checks for: Auto DJ (picker + `toggleAutoDJ`), star rating, the
-lyrics badge, share, and the torrent panel. `manage_server.dart` offers
-federated rows nothing but "Forget" — no edit, no pairing code, no storage
-settings.
+Every surface whose route the federation allowlist refuses is gated on
+`isFederated`, at the track level where a peer's track can sit in a queue
+browsed from another server:
 
-**`.m3u` rows.** The webapp hides them on a peer (mStream 9dfc3bfb): its row
-handlers called local APIs with a peer path, which poisoned the breadcrumb or
-loaded a same-named *local* playlist. The app's equivalents thread the server
-through (`makeServerCall(useThisServer, …)`) and `POST
-/api/v1/file-explorer/m3u` is allowlisted, so it may well work — but this is
-the one place the webapp found a path-namespace collision, so verify it on a
-peer before deciding to keep the rows.
+- **Auto DJ.** The panel's server dropdown skips peers; the queue-header
+  toggle refuses a peer with a snack ("Auto DJ can't run on a shared
+  server"); CarPlay's toggle and the car's Shuffle All refuse with a log
+  line; and the handler's `setAutoDJ` action is the backstop for every entry
+  point. A peer cannot host the DJ — random-songs is off the allowlist and
+  its paths mean nothing to the parent.
+- **Track sheet.** No rating badge and no lyrics badge for a peer's track,
+  and no Add to playlist (the parent would store a path it cannot resolve).
+  Download stays. The metadata screen's lyrics chip is gated the same way.
+- **Share** blocks a queue of peer tracks with its own message, before the
+  iroh "no public URL" block.
+- **Torrent.** The add-torrent screen offers only the user's own servers.
+- **Car root.** A peer's root has no Shuffle All and no Playlists
+  (`AutoBrowse.rootTabs`, unit-tested).
+- **Manage servers.** A peer row shows its name, a hub icon and "via
+  <parent>" (or "No longer shared by <parent>"), and offers Info plus
+  Hide/Show; no Edit, no pairing code. **The Forget decision:** peers are the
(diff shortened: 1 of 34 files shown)


IrosTheBeggar:
## Edge-case batch (2026-09-02, rig HTTP mode + one Quick Connect run)

Five lifecycle scenarios plus the cheap taps, run by hand on the kept rig. Three real gaps turned up and are fixed here (`e2c3dec`, `57c359c`); the recipe is in `smoke/recipes/federation-lifecycle.md`.

| Scenario | Result |
|---|---|
| Key revoked on the peer | browse fails fast (`db/album-songs → 502` in 54 ms, "Failed To Connect To Server"); a queued track falls back to its download; a failed auto-download raises the generic toast |
| Peer removed and re-added (new row id) | **fixed**: the old record adopts the new id by name (`peer id 1 → 2 (re-added on rig-b)`), one entry, downloads intact — it used to go missing and mint a `-2` twin |
| Parent's last peer removed while browsed | **fixed twice**: the parent's `federationBrowse` flag turns off with its last peer, so the reconcile never ran and both relaunches opened on the vanished peer → the flag going off now reconciles against an empty list; and launch / the removal fallback take the first *selectable* server. Now: `no longer shared by rig-b — leaving it for the parent`, second relaunch opens on the parent; Manage Servers shows "No longer shared by …", Info + Forget; Forget drops the record and its 6 queued tracks, keeps the folder unless ticked |
| Peer offline, parent up | `album-songs → 502 (15108ms)` after the parent's dial deadline, then the error state; the parent's bridge recovers on its own when the peer returns |
| Search / Artists / Recent / offline play | Artists and Recent 200 through the proxy; a downloaded peer track plays in airplane mode. **Search was unreachable on a peer** — the note row demoted the home to a plain list without the whole-server field — **fixed**: `isHomeList` treats a leading note as home, rendered as a banner above the card grid; `db/search → 200` returns "Ghosts" now |
| Parent outage over Quick Connect, peer track streaming | buffer runs out mid-outage → `iroh recovery: tunnel not ready for peer-rig-peer-a — parked`; on reconnect `iroh tunnel back — resuming parked playback`, PLAYING within 1 s (the Phase 4 transport path, live) |

Also found and fixed: tapping a peer's **row** in Manage Servers opened the Edit Server form (synthetic URL, meaningless toggles, a Save that would overwrite the record) — only the menu item had been hidden. The row opens Info now.

**Server-side notes for mStream** (not app issues): after a federation key is deleted the peer logs `usage flush failed for key id=1: FOREIGN KEY constraint failed` every 15 s; the per-endpoint backoff (5 failed handshakes → 60 s) blocks a peer that is re-added right after a revocation; and a just-enabled Quick Connect endpoint stalled the phone's first dials in the handshake (attempts 1–5, then success) in several runs.

`flutter test` 476, analyzer baseline.

Issue #143, opened 2026-09-03
Federation: connect the app directly to federated peers (guest tokens over the peer's federation endpoint)


IrosTheBeggar:
Handoff for the next agent. Everything below was established on 2026-09-02 while landing federation through the parent's proxies (#129, #141, #142). The goal of this work: let the app reach a **federated peer directly** — its bytes no longer cross the parent's home link twice, the peer stays usable while the parent is down, one hop less latency.

## Where things stand

- **Merged (proxy path):** a federated peer is a virtual `Server` reached through its parent's browse/art/stream proxies (`/api/v1/federation/peers/:id/{api,art,stream}`), with the parent's JWT. Phase 3 (capability gates) and Phase 4 (the identity/transport split) are done; `FEDERATION_PLAN.md` records the design and what each phase changed. Verified on a live two-server rig, including a lifecycle batch (`smoke/recipes/federation-lifecycle.md`).
- **Not started:** anything direct. **#115** (native tunnels keyed by pairing code, ABI v2 — "multi-iroh groundwork") is the only prior work in that direction; it is ~180 commits behind master and the Phase 2 supervisor rewrote the same files (`rust/iroh_tunnel/src/{ffi.rs,c_api.rs}`, `lib/native/iroh_tunnel.dart`), so treat it as a design reference, not a rebase.

## Why the app cannot dial a peer today (server trust model)

mStream federation is **server-to-server** over a dedicated iroh endpoint, ALPN `mstream/federation/1` (`src/state/federation.js`). The `mstrfed1:` ticket carries the peer's EndpointTicket `t` and a `fedk_` key `k`. The parent sends the raw key on the first bi-stream; the peer **TOFU-binds the key to the dialer's EndpointId** (`authenticateConnection`) — any other endpoint presenting the same key is rejected and logged as a leaked ticket. `docs/federation-ticket.md` states the key never leaves the parent. Every later bi-stream is a plain TCP-over-QUIC bridge into the peer's HTTP server; requests carry `x-federation-key`; a read-route allowlist (`src/api/federation-auth.js`) and per-key caps (`src/api/federation-limits.js`: maxStreams / dailyMb / streamKbps) apply. The parent's user-facing peer list (`GET /api/v1/federation/peers`) exposes only `{id, name, lastSeen, lastStatus, useDiscovery}` — **no endpoint id**.

So: no version of this works without server changes, and the parent's key must never reach a phone.

## Proposed design: guest tokens

[...]

PR #148, opened 2026-09-05
feat(iroh): keyed native tunnels, federation guest dial mode, in-place credential swap (ABI v2)

IrosTheBeggar:
App native half of #143 (server half: [IrosTheBeggar/mStream#943](https://github.com/IrosTheBeggar/mStream/pull/943)). **App behavior is unchanged in this PR** — the manager still runs one tunnel at a time and the one-Quick-Connect-server cap stays; this lands the native capabilities the multi-tunnel manager and the direct federated transport will build on.

## Rust (`rust/iroh_tunnel`)

- **Federation guest dial mode.** `parse_pairing_code` understands the `mstrfedg1:{t, g}` guest ticket next to the `mstr1:{t, s}` pairing code: ALPN `mstream/federation/1`, the guest token bytes (≤ 2 KB, the peer's `HANDSHAKE_LIMIT`) on the first bi-stream, `OK`/`NO` as before. A `mstrfed1:` federation ticket (the admin-to-admin one that carries a standing key) is refused by name. The three envelopes are pinned disjoint.
- **Credential swap in place.** `Tunnel::set_credential(code)` replaces the payload (and the endpoint address, which a refreshed ticket may update) without touching the loopback port or token. Connected: applies at the next re-dial. Reconnecting: the backoff is cut short, and a dial that was in flight with the stale credential and got rejected retries once with the new one. Rejected (supervisor exited): a fresh supervisor is spawned and re-dials immediately — the listener never left the port. Same kind and same endpoint id are enforced.
- The reconnect supervisor snapshots the credential per attempt; a guest-mode rejection logs "refresh it from the parent" instead of "re-pair needed"; events carry `mode=tunnel|guest`.
- **Keyed tunnels (ABI v2).** `ffi.rs` holds a `BTreeMap<key, Tunnel>` — the key is an app-chosen id (the server's identity), deliberately not the credential, so a refreshed token does not mean a new tunnel. Every per-tunnel entry point takes the key; `network_changed` fans out to all. `c_api.rs` renames nothing but changes signatures, adds `mstream_iroh_abi_version` (2) and `mstream_iroh_set_credential`; 14 exported symbols.
- Dev CLI: `--swap-after <secs> <code>`. Interop harness: a GUEST phase (a JS federation-ALPN endpoint; a refused token fails the start with "rejected"; the guest dial serves HTTP; the server flips to a new token and drops the pipe so the re-dial is refused; the client's timed swap respawns the supervisor on the same port).

## Dart

- `lib/native/iroh_tunnel.dart` is keyed throughout (`start(key, code)`, `stop(key)`, `statusOf`, `pathKindOf`, `localTokenOf`, `kick(key)`, `drainEvents(key)`, `relayOnlineOf`, `setCredential(key, code)`). It probes the ABI version first: a stale binary makes `isSupported` false with the reason in `unsupportedReason` rather than being driven with arguments it would misread.
- `ServerManager` keys its single tunnel by the pairing code it dialed (`_activeTunnelCode` is now also the native key) through four small accessors; `add_server.dart` remembers the code each test dial used (`_irohTestedCode`) so a later stop targets the right tunnel even after the text field changed.

## Binaries

Both rebuilt and committed (`build-android.sh` arm64-v8a + x86_64, `build-ios.sh` device + simulator); the iOS script's export check expects 14 symbols; README checksums updated.

## Verification

- `cargo test`: 14 (guest ticket parse, limits, disjointness, refusal of federation tickets)
- `node interop/harness.mjs`: ALL PASS — interop, reconnect, kick, and the new guest phase
- `flutter analyze`: no findings in the touched files (repo baseline unchanged); `flutter test`: 480 pass
- Android (arm64 emulator, debug `.plus.dev` build with the new `.so`, the interop `pairing-server.mjs` as the endpoint): `nativeloader` loads the library, `[iroh] tunnel up port=32861 path=direct in 1095ms (launch)`, native events drain through the keyed call (`mode=tunnel`), and the network-change probe returns `http 200` through the tunnel
- iOS (iPhone 17 Pro simulator, same server): the rebuilt framework exports 14 symbols inside the app, `[iroh] tunnel up port=53132 path=direct in 3189ms (launch)`

## Next (per #143)

Multi-tunnel `ServerManager` (per-server handles, the queue listener as a set, banner keyed to the browsed transport), then the direct transport on the federated `Server` with the proxy as fallback and the daily token refresh through `setCredential`, then the rig in direct mode.


IrosTheBeggar:
## Smoke run on the Galaxy S25 (2026-09-04, this branch's debug `.plus.dev` build, personal server paired both ways)

| Script | Result | Notes |
|---|---|---|
| `session-actions` | 2/2 | |
| `launch-matrix` | 6/6 | standard default published in 0.5 s; Quick Connect default: tunnel up in 3.8 s, one dial, no teardown churn |
| `switch-during-outage` | 2 pass, 1 skip | reconnecting edge + switch logged 0.1 s after the tap; the "tunnel back" check skips because the saved queue held only standard-server tracks, so the switch released the tunnel (the script now says so instead of failing) |
| `media-resumption` | 3/3 | headless boot on a PLAY key, queue restored |
| `dead-zone` | 6 pass, 1 fail | hand-off kept (probe passed); outage 1 reconnected in place +22.4 s; outage 2 reconnected in place +29.1 s; Wi-Fi return serving +23 s via the in-place kick; no unexplained rebuild. The fail is the Retry-tap step: the tunnel was still reporting *connected* 95 s into outage 2, so there was no banner to tap — see below |
| `federation-rig` (Quick Connect mode, servers from mStream#943) | 11/11 | peer reconciled under the tunnel parent in 28 s, browsed, played through the stream proxy, download folder created, tunnel kept while browsing the peer and after switching to a standard server, restored peer track plays after a relaunch (port 40233 → 42753) |
| iOS `sim-launch` | 3/3 | with a Quick Connect default; tunnel up in 3.2 s on the simulator |

**The dead-zone Retry step.** After outage 1 the supervisor reconnected on a *direct* path with the relay offline; when outage 2 began 10 s later, iroh did not report the connection closed for 138 s (the QUIC "timed out" close arrived 13 s after service returned). Relay-path connections flip to reconnecting in 32–36 s in every run on disk. The same pattern is in the 2026-09-02 run `20260902-203200` on the previous binary (outage 2 on a direct path: no edge inside the window), so it is pre-existing iroh behaviour rather than something this PR changed; recorded for the tunnel-policy follow-ups (a probe on the `[none]` edge would surface it).

Smoke-suite fixes in the second commit: `applog` is scoped to the dev package's uid (the production build's tunnel lines were leaking into the checks), and the switch script reports the released-tunnel case as a SKIP with the reason.

PR #149, opened 2026-09-05
feat(iroh): one tunnel per transport server — the multi-tunnel ServerManager

IrosTheBeggar:
Step 3 of #143: `ServerManager` runs one native tunnel **per transport server** instead of one for the whole app. Groundwork for the direct federated transport (next), which needs a peer's own tunnel next to its parent's. **App behavior for a single Quick Connect server is unchanged**, and the one-Quick-Connect-server cap in the add-server screen stays.

## What changed

- **`TunnelHandle`** (`lib/singletons/tunnel_handle.dart`, new): one native tunnel's bookkeeping per transport server — its dial chain, retry ladder, probe/kick/watchdog state, the port and token it mirrors onto the `Server`, and the native key it was started under (a re-pair swaps the server's code before the old tunnel is stopped, so the key is recorded, not recomputed). Pure; unit-tested.
- **The manager** keeps a table of handles keyed by transport localname and runs the same lifecycle on each. Targets are the browsed server's transport when it is a tunnel plus every tunnel-carried transport the play queue references. `ensureTunnels` reconciles handles against the targets (start the missing, leave the healthy, release the unreferenced); `ensureTunnelFor` (re)starts one server's tunnel on its own chain, which is what the playback, download and car paths ask for through `awaitTunnelReady`. One server's cold dial never holds another's.
- **Queue listener** reports the whole set of tunnel-carried transports (`setQueueIrohServers`); each transport the queue lets go of is released after `queueReleaseGrace`, exactly as the single pointer was.
- **Status strip**: one value, the browsed server's tunnel when it is one, else the background playback tunnel in the worst state (`bannerTargetAmong`, unit-tested); `TunnelPolicy.showTunnelBanner` still decides whether that state deserves a strip. `handleNetworkChange` runs every target's probe round concurrently; a Retry tap applies its user semantics to the strip's tunnel. The manage-servers chip reads its own server's tunnel (`tunnelStatusOf` / `pathKindOf`).
- **Playback heal** listens to `tunnelTransitions` (per tunnel, tagged with its transport) instead of the strip's single value, so one tunnel's reconnect edge can never hide behind another's state.
- **Log lines** keep their shapes (the smoke scripts grep them) and gain a `for=<localname>` suffix.
- The cast fallback on "switching to a different iroh server tears down the only tunnel" is gone with the reason for it: selecting another server no longer tears a tunnel out from under a cast; only the queue letting go of it does, after the grace.

## Verification

- `flutter analyze` at the repo baseline; `flutter test` 488 (8 new: key choice, bind/clear mirroring, identity re-check across a code swap, banner-target selection, severity order)
- Emulator (arm64 AVD, local mStream 6.25 with Quick Connect on, debug build):
  | Script | Result |
  |---|---|
  | `launch-matrix` | 6/6 — standard default published in 0.5 s; Quick Connect default: tunnel up in 1.6 s, one dial, no teardown churn |
  | `dead-zone` | 7/7 — hand-off kept, outage 1 reconnected in place +1.7 s, Retry kicked in place, outage 2 in place +13.2 s, Wi-Fi return serving +9.9 s, no unexplained rebuild |
  | `switch-during-outage` | 3/3 — with Quick Connect tracks queued, the tunnel survived the switch and was back 7 s after the outage ended |
  | `federation-rig` (Quick Connect mode) | rig steps 9/11 with the Galaxy tap coordinates (the two "fails" were taps landing between tiles on the Pixel 7 layout); the peer playback half driven by hand with the emulator's coordinates: album queued and **PLAYING through the parent's tunnel** (`127.0.0.1:<parent port>/api/v1/federation/peers/1/stream/…`), the Quick Connect tunnel the queue no longer referenced **released after the grace while the parent's stayed** (`queue no longer references iroh-smoke — releasing its tunnel` / `tunnel stopped (no-target/queue-server) … for=iroh-smoke`, no stop for `iroh-rig-b`), switch to the standard server mid-track: still playing, parent tunnel kept; relaunch: parent tunnel up on a fresh port (36845 → 44283), the restored peer track plays with its URL rebuilt onto that port |

  The relaunch during the rig showed two tunnels coming up side by side (the browsed parent at launch, the queue's Quick Connect server 0.4 s later) — the first time the app has run two.

## Next (per #143)

The direct transport on the federated `Server`: an access call to the parent, a handle keyed by the peer's localname started from the guest ticket, the proxy path as fallback, and the daily token refresh through `IrohTunnel.setCredential` — then the two-server rig in direct mode.

### DIFF
--- a/FEDERATION_PLAN.md
+++ b/FEDERATION_PLAN.md
@@ -275,6 +275,44 @@ the queue-restore re-origin. Casting a peer track through the LAN relay is
 covered by the same code path as an iroh server's and is the one piece
 not run on a renderer.
 
+### Phase 6 — one tunnel per transport server ✅ done
+
+Groundwork for reaching a peer directly (issue #143): the direct transport
+needs a peer's own tunnel next to its parent's, and `ServerManager` was
+single-tunnel throughout — one pairing code, one dial chain, one retry
+ladder, one probe/kick/watchdog bookkeeping, a queue listener that picked
+the first tunnel-carried track, a status strip fed by one value.
+
+**The model.** [`TunnelHandle`](lib/singletons/tunnel_handle.dart) holds one
+native tunnel's bookkeeping per *transport* server (a Quick Connect server
+today; a directly-reached peer next), keyed by localname in the manager's
+`_tunnels` table. Each handle owns its own chain, retry ladder, probe and
+watchdog state, so a second tunnel is a second entry, not a second copy of
+the lifecycle — and one server's cold dial never holds another's. The
+native key a handle was started under is recorded on it (a re-pair swaps
+the server's code before the old tunnel is stopped).
+
+**Targets.** The browsed server's transport when that is a tunnel, plus
+every tunnel-carried transport the play queue references — the queue
+listener now reports the whole set (`setQueueIrohServers`), and each
+transport it lets go of is released after `queueReleaseGrace`. `ensureTunnels`
+reconciles handles against the targets; `ensureTunnelFor` (re)starts one
+server's tunnel on demand, which is what the playback, download and car
+paths ask for.
+
+**What stays single.** The status strip shows one value: the browsed
+server's tunnel when it is one, else the background playback tunnel in the
+worst state (`bannerTargetAmong`, unit-tested) — `TunnelPolicy.showTunnelBanner`
+still decides whether that state deserves a strip. Listeners that care about
+one server's tunnel (the playback heal watches the parked track's) take
+`tunnelTransitions`, so one tunnel's edge can never hide behind another's
+state. The one-Quick-Connect-server cap in the add-server screen stays for
+now; lifting it is a separate change.
+
+**Log lines** keep their shapes (the smoke scripts grep them) and gain a
+`for=<localname>` suffix, since two tunnels would otherwise be
+indistinguishable in a Diagnostics export.
+
 ### Phase 5 — optional: make Discover leads actionable
 
 `/api/v1/discovery/federation/similar` already returns `peer:{id,name}` on the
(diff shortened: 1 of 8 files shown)


IrosTheBeggar:
## Smoke round on the Galaxy S25 (2026-09-04, this branch's debug `.plus.dev` build, personal server paired both ways, real cellular)

| Script | Result | Notes |
|---|---|---|
| `session-actions` | 2/2 | |
| `launch-matrix` | 6/6 | standard default in 0.5 s; Quick Connect default: tunnel up in 3.8 s, one dial, no teardown churn |
| `dead-zone` | 7/7 | hand-off kept; outage 1 reconnected in place +28.7 s; **Retry kicked in place**; outage 2 in place +18.6 s; Wi-Fi return serving +23.2 s via the in-place kick; no unexplained rebuild |
| `switch-during-outage` | 3/3 | with Quick Connect tracks queued: reconnecting edge, switch logged 0.1 s after the tap, tunnel back 7 s after the outage ended |
| `media-resumption` | 3/3 | headless boot on a PLAY key, queue restored |
| `federation-rig` (Quick Connect mode, servers from mStream#943) | 11/11 | peer reconciled under the tunnel parent in 4 s, browsed, played through the stream proxy, tunnel kept while browsing the peer and after switching to a standard server, restored peer track plays after a relaunch on the fresh port (37371 → 43341) — the relaunch check now verifies the restored item is the peer's on that port |

**Multi-tunnel in the wild.** The phone's queue held 521 Quick Connect tracks going into the rig, so the app launched with two tunnels (the browsed rig parent, plus the queue's server). When the peer album replaced the queue, the manager released the Quick Connect tunnel after the grace and kept the parent's — playback never paused. The rig's first pass read that release as "tunnel lost" because it counted any `tunnel stopped` line; the check is now scoped to the parent (third commit).

**Smoke-suite fixes on this branch:** `queue.json` travels with the device-config backup (the rig used to leave the phone on a queue of peer tracks whose server it had removed — the next PLAY key did nothing), the rig's relaunch check reads the restored item's server and port, its tunnel-stop checks are parent-scoped, and the emulator tap coordinates are documented in the script header.

PR #150, opened 2026-09-05
feat(federation): direct access to a peer over its own tunnel (#143 step 4) + Quick Connect cap removal

IrosTheBeggar:
Step 4 of #143, with the one-Quick-Connect-server cap removal folded in. Server side is mStream#943 (merged); native layer #148; multi-tunnel manager #149.

## What changes

A federated peer whose parent advertises direct access (`federationDirect` in the parent's `/api` user block) now gets a tunnel of its own. The manager asks the parent for the peer's guest ticket (`GET /api/v1/federation/peers/:id/access`), dials the peer's federation endpoint with it (the `mstrfedg1:` ticket, `mode=guest` on the native side), and the peer's URLs move to its own loopback with the guest token: `/media` and `/album-art` straight from the peer, `apiUri` the plain path. No proxy hop, and the peer keeps playing and browsing while the parent is down. The proxy path stays the fallback for an older peer, a declined mint, or a failed dial.

**The mode is the peer's own tunnel port.** `Server.isDirect` is true exactly while the peer holds a `tunnelPort` of its own (the manager binds and clears it with the tunnel), and every accessor reads through it. `ownsTunnel` (a Quick Connect server, or a direct peer) is the "has a tunnel of its own" question the manager and the strip ask; `isIroh` stays identity (the pairing menu, the repair sheet).

**Manager.** A browsed or queued peer is a tunnel target of its own when its parent offers direct access, nobody declined this session, and the parent still lists it; the parent's tunnel stays a target while the peer is not direct yet and whenever the peer's ticket has to come through it again. The ticket fetch sits behind the existing backoff gates (a failing parent is asked on the retry ladder, not on every reconcile). The queue listener now reports item servers rather than transports, since a peer's transport changes with its mode.

**Ticket lifecycle.** Fetched on demand; renewed in place once 75 % of its life is gone (`IrohTunnel.setCredential`, same port, the queued URLs survive; only upcoming items take the new token). A stale-but-unchanged answer forces the parent to re-mint. Only failed attempts are rate-limited (5 min), an expired ticket retries on the 60 s gap, and a parent that comes back after an outage is asked within a minute. A refused guest token is asked for again rather than treated as a re-pair: at the first dial through one forced refresh and a quick retry, on a running tunnel through the poll, on a browse 401 at once. `direct: false` puts the peer on the proxy for the session. When a direct tunnel goes, the queued URLs are rebuilt against the parent. A peer handle that is still an attempt never puts Repair/Retry on the strip (the proxy is serving).

**Cap removal.** The native layer keys tunnels by server and the manager runs one per server, so the add-server screen and its backstop no longer refuse a second Quick Connect server. The unused l10n string is gone too.

## Tests

504 pass (16 new): `direct_access_test.dart` (parse ok/declined/malformed, staleness boundaries), `stream_url_direct_test.dart` (both URL shapes and the flip), `federated_server_test.dart` (direct-mode accessors, runtime-only fields, the persisted parent flag), `tunnel_handle_test.dart` (`credentialFor`, bind/clear flips the mode, the strip follows a direct peer).

## Verified on the Galaxy S25 (two-server rig, `MSTREAM_TEST_FED_GUEST_TTL_MS=180000`, parent over LAN HTTP)

- Ticket issued 70 ms after the switch to the peer; peer tunnel up (`path=direct`, `mode=guest`) after one 13 s handshake stall on the first dial; upcoming queue items rebuilt onto `127.0.0.1:<peer port>/media/...` with the guest token and the peer's own `__lt`.
- Three renewals in place at 75 % (the third through a parent restarted with an empty cache), one `tunnel up` for the whole session: port kept.
- Parent process killed: a freshly queued peer album streams from the peer's loopback (`state → ready` in 0.2 s, PLAYING), and browsing back into the album answers `POST /api/v1/file-explorer → 200` over the direct tunnel; the peer logged zero rejected tokens.
- The existing rig (`federation-rig.sh`) passes 7/7 twice on this build.

## Observations (not regressions, not fixed here)

- The first guest dial stalled in the handshake for 10 s once or twice before landing (the retry ladder covered it): the same first-dial stall as mStream#940, now on the federation endpoint. The peer logs nothing for it.
- The keep-queue-offline downloader starts every track of a queued album at once; a federated peer caps a key at 3 concurrent streams (guests share it), so the rest get 429 and retry later, and a burst can compete with the player's own stream. Predates this PR (the proxy path counts under the same key).
- A downloaded item whose local copy vanishes behind the app's back falls back to its stored URL, which can carry an expired guest token (rebuilds skip downloaded items). Re-deriving the URL at load time would close it; same class as a stale `__lt` after a tunnel rebuild.

## Next

Step 5: a direct-mode flag for `smoke/android/federation-rig.sh` (parent killed mid-track, key revoked, short TTL renewal), replacing the hand-run script used above.

### DIFF
--- a/FEDERATION_PLAN.md
+++ b/FEDERATION_PLAN.md
@@ -313,6 +313,54 @@ now; lifting it is a separate change.
 `for=<localname>` suffix, since two tunnels would otherwise be
 indistinguishable in a Diagnostics export.
 
+### Phase 7 — direct access: a peer over a tunnel of its own ✅ done
+
+The point of issue #143: a peer's bytes no longer cross the parent's home
+link twice, the peer stays usable while the parent is down, and there is one
+hop less latency. Server side: mStream#943 (guest tokens, the parent's
+`access` route, the `mstrfedg1:` guest ticket, `federationDirect` in the
+`/api` user block). App side, this phase.
+
+**The mode is the peer's own tunnel port.** `Server.isDirect` is true while
+the peer holds a `tunnelPort` of its own — the manager binds it when the
+peer's tunnel is up and clears it when that tunnel goes — and every accessor
+reads through it: `transportServer` is the peer itself, `effectiveBaseUrl`
+its loopback, `authToken` the guest token, `localTokenQuery` its own `__lt`,
+`apiUri` the plain path. The three URL builders take the plain `/media` and
+`/album-art` shapes for a direct peer and the parent's proxy shapes
+otherwise; `albumArtFileFromUrl` already read both. `ownsTunnel` (a Quick
+Connect server, or a direct peer) is the "has a tunnel of its own" question
+the manager and the strip ask; `isIroh` stays identity.
+
+**The manager.** A federated peer that is browsed or queued is a target of
+its own when its parent advertises `federationDirect`, nobody declined this
+session, and the parent still lists it — with the parent's tunnel kept as a
+target too until the peer is direct (the proxy serves meanwhile). The peer's
+handle keys by localname and dials with its guest ticket
+(`TunnelHandle.credentialFor`), fetched from the parent on demand
+(`GET …/peers/:id/access`, waiting for a Quick Connect parent's tunnel
+first) and recorded on the peer with its expiry. A tunnel that is up gets a
+fresh ticket swapped in place once three quarters of the lifetime is gone
+(`IrohTunnel.setCredential` — same port, the queued URLs survive; only
+upcoming items take the new token). A refused guest token is asked for
+again rather than treated as a re-pair: at the first dial through one forced
+refresh and a quick retry, on a running tunnel through the poll, on a browse
+401 at once. A parent that answers `direct: false` puts the peer on the
+proxy for the session. When a direct tunnel goes, the queued URLs are
+rebuilt against the parent.
+
+**What the user sees.** The strip follows a browsed direct peer's own
+tunnel; a peer handle that is still an attempt never puts Repair or Retry on
+the strip (the proxy is serving). Log lines: `[federation] <peer>: direct
+access issued/unchanged/no direct access`, the peer's own `[iroh] … for=<peer>`
+lifecycle with `mode=guest` in the native events.
+
+**Also in this phase: the one-Quick-Connect-server cap is gone.** The native
+layer keys tunnels by server and the manager runs one per server, so the
+add-server screen and its backstop no longer refuse a second Quick Connect
+server. A server that is neither browsed nor queued has no tunnel, so a phone
+with several paired servers still runs one most of the time.
+
 ### Phase 5 — optional: make Discover leads actionable
 
 `/api/v1/discovery/federation/similar` already returns `peer:{id,name}` on the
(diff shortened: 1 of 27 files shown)


IrosTheBeggar:
Follow-up commit on this branch: two gaps found while verifying the renewal path on the Galaxy.

**Fixes**
- An upcoming-only rebuild now widens to the current track when that track is idle or still loading (`AudioPlayerHandler.protectsCurrentTrack`). Before, a current track failing on an expired guest token kept its stale URL through the renewal, and the player re-seeded it forever.
- Rebuilds no longer skip a downloaded item whose file is gone (`hasLocalCopy`); its URL is refreshed like a streaming item's.

**Galaxy verification** (rig, HTTP parent, `MSTREAM_TEST_FED_GUEST_TTL_MS=180000`, keep-queue-offline off so tracks stream)

| Step | Result |
|---|---|
| Parent killed before the first renewal, token expired, `NEXT` | load refused, failure loop running (state ERROR) |
| Parent restarted | renewal 61 s later (expired ticket → 60 s gap) |
| Rebuild | `upcoming-only rebuild widened to the current track (idle)`, reload at the same track |
| Playback | playing again 1 s after the reload, no failures afterwards |

Also re-run on this build: the rig 7/7, the direct-mode pass (peer tunnel up `mode=guest`, three in-place renewals with the port kept, fresh queue streams + `POST /api/v1/file-explorer → 200` with the parent process dead).

**Seen, not fixed here:** while the token was expired the heal loop re-seeded the same track every ~14 s and ignored its own "tunnel probe passed — the failures were not the path" verdict. That, plus the download burst against a peer's 3-stream cap, is the follow-up resilience PR.

IrosTheBeggar:
Overnight release verification of the whole stack (#150 → #151 → #152) is posted on #151: the full Android smoke suite green on the release candidate (both federation rig modes incl. renewal and a revoked key, heal hand-off, download cap, launch matrix, switch, media resumption, dead zone), the iOS simulator's direct round green twice, the iPhone waiting on Developer Mode. Merge order: this PR first, then #151, then #152.

IrosTheBeggar:
**iPhone X (iOS 16.7.16), Developer Mode on, debug build under the debugger, restored peer queue**

| Step | Result |
|---|---|
| Launch with a 6-track peer queue restored | guest ticket issued, the peer's own tunnel up on the first dial (3.7 s, `path=direct`, `mode=guest`) |
| Queue at launch (restored, paused) | `upcoming-only rebuild widened to the current track (ready, paused)`; all 6 URLs on `127.0.0.1:<peer port>` with the guest token and the peer's `__lt` |
| Renewal at 75 % of a 3-minute ticket | `guest credential refreshed in place`, same port |
| Parent killed, restarted 30 s later | next renewal through the restarted parent, same port |

So the native guest dial and the in-place credential swap work on the iOS xcframework on hardware, not only in the simulator. Playback itself was not exercised on the device (no input injection on a physical iPhone).

Gotcha for next time: ios-deploy needs `~/Library/Developer/Xcode/iOS DeviceSupport/<model> <version> (<build>)/Symbols` to exist for this device before it will launch anything; Xcode creates it when it first prepares the phone, or `mkdir -p` does.

PR #151, opened 2026-09-05
fix(playback): the tunnel heal gives up on a verified path; downloads run two per server

IrosTheBeggar:
Stacked on #150 (the two observations from its verification that were not fixed there). Both were found on the two-server rig with an expired guest token, but neither is specific to federation: a Quick Connect server whose token died, or a queued album on any small server, hits the same paths.

## 1. The tunnel heal loop now terminates

When every load fails against a tunnel that is up (an expired or rotated credential, files the server no longer has), the heal re-seeded the same track every ~14 s for as long as a serving tunnel and an idle player coexisted. Its own ground-truth probe said "the failures were not the path" and nothing acted on it, and the ordinary retry-then-skip walk never got a turn because every park re-triggered the heal.

- `ServerManager.reverifyTunnel` returns its verdict (true = the probe ran and passed).
- After a failed heal resume, `healResumeFailureAction` decides: re-arm while the path may still be the problem; hand the track to the skip walk once the probe has cleared the path, and after six failed resumes regardless (a probe that never ran cannot keep the loop alive).
- A hand-off suspends the heal (`healAction` → drop) until a track loads or a real tunnel edge arrives, so the walk runs at the ordinary pace instead of two heal cycles per track.

## 2. Keep-queue-offline downloads two at a time per server

A queued album fired every track at once. A federated peer caps a key at 3 concurrent streams (guests share it), so most came back 429 to retry in another burst, and the burst competed with the player's own stream; a small self-hosted server is not happier about 26 range requests either.

- `autoDownloadCandidates` takes a per-server slot budget and marks only what it returns; the rest stay unmarked and enqueue as transfers land (a landing re-sweeps).
- `kAutoDownloadParallel = 2`; manual downloads are not held back.
- Sweeps are serialized: they fire on every queue emission, and two running at once each read the in-flight set before the other's enqueues landed (measured: three transfers at a queue's birth → eight 429s in the first seven seconds, none afterwards).

## Tests

514 pass (6 new): `healResumeFailureAction`, the suspended heal, the slot budget and its marking.

## Galaxy S25 verification

Two-server rig, 3-minute guest TTL, keep-queue-offline off unless noted. The full release suite (both rig modes, launch matrix, switch, media resumption, dead zone) is running on the phone overnight; its table follows as a comment.

| Scenario | Result |
|---|---|
| Expired token, parent dead: 3 failed resumes → probe clears the path → hand-off | PASS, heal stays out of the rest of the walk |
| Same, before the retry fix | the walk sat at retry 1 (found and fixed in this PR) |
| Parent back: renewal → widened rebuild → playing again | PASS, 1 s after the renewal |
| Queued album with keep-queue-offline on | 0 stream-cap rejections at the queue's birth (was 23, then 8 before the sweeps were serialized), 26/26 landed |
| Key revoked on the peer (rig) | app alive 75 s later, 1 access line, 0 heal resumes, playback parked |
| iOS simulator, same rig | direct tunnel on the first dial, playback and browse with the parent dead, two in-place renewals |

### DIFF
--- a/lib/media/audio_stuff.dart
+++ b/lib/media/audio_stuff.dart
@@ -57,6 +57,10 @@ enum MediaToggleAction { play, pause, ignore }
 /// this trigger entirely.
 enum HealAction { run, rearm, drop }
 
+/// What to do after a tunnel-heal resume failed: try again later, or stop
+/// blaming the tunnel and let the failing track take the ordinary skip walk.
+enum ResumeFailureAction { rearm, skipTrack }
+
 class AudioPlayerHandler extends BaseAudioHandler
     with QueueHandler, SeekHandler {
   // ignore: close_sinks
@@ -354,7 +358,11 @@ class AudioPlayerHandler extends BaseAudioHandler
     // one's state, and that edge must not be lost.
     ServerManager().tunnelTransitions.listen((t) {
       if (t.from == IrohTunnelStatus.connected) return;
-      if (t.to == IrohTunnelStatus.connected) _onTunnelReconnected();
+      if (t.to == IrohTunnelStatus.connected) {
+        // A real edge: whatever verdict suspended the heal is stale now.
+        _healSuspended = false;
+        _onTunnelReconnected();
+      }
     });
     // Stop the service when playback reaches the end of the queue.
     _backendSubject.switchMap((b) => b.processingStateStream).listen((state) {
@@ -428,6 +436,8 @@ class AudioPlayerHandler extends BaseAudioHandler
         _httpRetries = 0;
         _networkStalled = false;
         _tunnelLoadFailures = 0;
+        _healResumeFailures = 0;
+        _healSuspended = false;
         // Only a DIFFERENT track reaching ready clears the local-copy retry
         // guard. Clearing it on every ready reopens the loop it exists to
         // stop: a corrupt-but-LOADABLE file (valid header, damaged tail)
@@ -1033,7 +1043,7 @@ class AudioPlayerHandler extends BaseAudioHandler
     if (itemServer != null &&
         itemServer.isIrohTransport &&
         ServerManager().tunnelServes(itemServer)) {
-      _noteTunnelLoadFailure(itemServer);
+      unawaited(_noteTunnelLoadFailure(itemServer));
     }
     // Non-iroh source failed → hand off to _recoverHttpError, which (with an
     // async connectivity probe) pauses-and-holds on a real outage, retries a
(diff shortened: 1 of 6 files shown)


IrosTheBeggar:
Overnight release verification on the stack #150 → #151 → #152 (Galaxy S25, debug `.plus.dev` build of the head of #152; two-server rig, 3-minute guest TTL unless noted).

**Android smoke suite, all on the release candidate**

| Run | Result |
|---|---|
| Federation rig, HTTP parent (proxy checks + direct phase + renewal + key revoked) | 16 / 16 |
| Federation rig, Quick Connect parent (+ parent released/re-dialed, switch, relaunch) | 21 / 21 on the third attempt — the first two never got the parent's fresh Quick Connect tunnel up (six 13 s handshake stalls in a row, mStream#940); a 15 s warm-up after enabling Quick Connect fixed the rig, not the app |
| Heal hand-off: expired token, parent dead → probe → skip walk, parent back → renewal → resume | all pass, walk at ~4 s/track ending cleanly |
| Download cap, keep-queue-offline on | 0 stream-cap rejections (was 23), 26/26 landed, playback undisturbed |
| launch-matrix | 7 / 7 |
| switch-during-outage | 3 / 3 |
| media-resumption | 3 / 3 |
| dead-zone (cellular, two airplane outages, Retry, Wi-Fi return) | 7 / 7, reconnects in place, no rebuilds |

**iOS**

| Run | Result |
|---|---|
| Simulator (iPhone 17 Pro), full direct round | ticket on the switch, own tunnel on the first dial (3.2 s, direct, guest), albums and album-songs 200 over it, 6 queued URLs on the peer's loopback, renewal in place, parent killed → next track + browse OK, parent restarted → second renewal; repeated on the final build |
| Simulator `sim-launch.sh` | boot + default published pass; the resume check timed out because the restored config's Quick Connect server isn't running (the resume line is in the log later) |
| iPhone X | now on iOS 16.7 with Developer Mode off, so nothing can be launched or attached from the Mac. It has a profile build of this stack installed and the rig parent planted first in its server list; the rig pair for it is left running (peer 3111, parent 3112) for a hand test in the morning. Backups of its documents are under `smoke/out/device-backups-20260905/`. |

**Fixed along the way (in #151 / #152)**

- Quick Connect parent: once a peer went direct, nothing reconciled the parent's tunnel away until the next queue edit or switch — a bind now reconciles, and a released tunnel rebuilds whatever still rode it.
- A queue restored at launch sits ready-but-paused; an upcoming-only rebuild now widens to a paused current track as well, so it moves to the peer's loopback instead of staying on a proxy port that may be released.
- The rig: the queued-URL check ignores downloaded copies and polls; a Quick Connect reconcile waits out the first-dial stall ladder; after revoking the key the rig presses NEXT so a buffered track cannot mask the outcome.

**Still open**

- The first guest dial through a freshly enabled Quick Connect endpoint stalls in the handshake for 13 s a time, up to four times tonight (mStream#940). The retry ladder lands it; the simulator never stalls.
- Casting a direct peer to a renderer was not run (a TV in the house at night).

IrosTheBeggar:
**iPhone X (iOS 16.7.16), Developer Mode on, debug build under the debugger, restored peer queue**

| Step | Result |
|---|---|
| Launch with a 6-track peer queue restored | guest ticket issued, the peer's own tunnel up on the first dial (3.7 s, `path=direct`, `mode=guest`) |
| Queue at launch (restored, paused) | `upcoming-only rebuild widened to the current track (ready, paused)`; all 6 URLs on `127.0.0.1:<peer port>` with the guest token and the peer's `__lt` |
| Renewal at 75 % of a 3-minute ticket | `guest credential refreshed in place`, same port |
| Parent killed, restarted 30 s later | next renewal through the restarted parent, same port |

So the native guest dial and the in-place credential swap work on the iOS xcframework on hardware, not only in the simulator. Playback itself was not exercised on the device (no input injection on a physical iPhone).

Gotcha for next time: ios-deploy needs `~/Library/Developer/Xcode/iOS DeviceSupport/<model> <version> (<build>)/Symbols` to exist for this device before it will launch anything; Xcode creates it when it first prepares the phone, or `mkdir -p` does.

IrosTheBeggar:
Landed in master through #152's merge (c2092efa): #152 was stacked on this branch, so its merge carried both of this PR's commits (6fbc865, 2884691). GitHub closed this PR when #150's branch was deleted and would not let it be retargeted afterwards; the code is in master regardless.

PR #152, opened 2026-09-05
smoke(rig): federation rig covers direct access, a Quick Connect parent, and a revoked key (#143 step 5)

IrosTheBeggar:
Issue #143 step 5, stacked on the resilience PR (which is stacked on #150).

`smoke/android/federation-rig.sh` now covers the direct path. After the existing proxy checks it detects direct access from the log (an older server is skipped, not failed) and checks:

- the guest ticket issued by the parent;
- the peer's own tunnel (`mode=guest`, port and path reported);
- the upcoming streaming URLs moved to the peer's loopback (`/media`, guest token, own `__lt`), none left on the proxy — downloaded copies keep their stored URL by design and are ignored;
- the parent killed mid-track: the next track streams and the peer's file explorer answers 200 over the direct tunnel;
- the ticket renewed in place at 75 % of `SMOKE_RIG_TTL_MS` (default 180000; 0 skips the wait), port kept;
- Quick Connect mode: the parent's tunnel released once the peer is direct, and re-dialed for the access call at renewal; the switch and relaunch checks follow whichever tunnel carries the peer;
- `SMOKE_RIG_REVOKE=1` (default) ends by revoking the key on the peer: the app must stay alive, not loop, and park.

`SMOKE_RIG_PA` / `SMOKE_RIG_PB` pick the ports, and `SMOKE_RIG_SERVERS_ONLY=1` starts and pairs the servers, prints their details as JSON and leaves them up with no phone attached — the iOS simulator round was driven by hand against a pair started this way while the Galaxy used the default ports.

First automated pass on the Galaxy (HTTP parent): 13 of 13 including the whole direct phase. The Quick Connect-parent pass and the revocation step are part of the overnight suite; results follow as a comment.

### DIFF
--- a/FEDERATION_PLAN.md
+++ b/FEDERATION_PLAN.md
@@ -361,6 +361,16 @@ add-server screen and its backstop no longer refuse a second Quick Connect
 server. A server that is neither browsed nor queued has no tunnel, so a phone
 with several paired servers still runs one most of the time.
 
+**Verification (2026-09-05).** `smoke/android/federation-rig.sh` covers the
+direct path (ticket, own tunnel, URLs moved, parent killed mid-track, renewal
+in place, a Quick Connect parent released and re-dialed, a revoked key) in
+both parent modes; `SMOKE_RIG_SERVERS_ONLY=1` starts a pair for the iOS
+rounds. Playback resilience found on the way (the tunnel heal handing a
+verified-path failure to the skip walk, two downloads per server) is its own
+PR. Galaxy S25: the whole Android smoke suite green on the release
+candidate; iOS simulator: the full direct round green; iPhone: needs
+Developer Mode for a launched test.
+
 ### Phase 5 — optional: make Discover leads actionable
 
 `/api/v1/discovery/federation/similar` already returns `peer:{id,name}` on the
--- a/lib/media/audio_stuff.dart
+++ b/lib/media/audio_stuff.dart
@@ -57,6 +57,10 @@ enum MediaToggleAction { play, pause, ignore }
 /// this trigger entirely.
 enum HealAction { run, rearm, drop }
 
+/// What to do after a tunnel-heal resume failed: try again later, or stop
+/// blaming the tunnel and let the failing track take the ordinary skip walk.
+enum ResumeFailureAction { rearm, skipTrack }
+
 class AudioPlayerHandler extends BaseAudioHandler
     with QueueHandler, SeekHandler {
   // ignore: close_sinks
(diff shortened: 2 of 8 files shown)


IrosTheBeggar:
Overnight release verification on the stack #150 → #151 → #152 (Galaxy S25, debug `.plus.dev` build of the head of #152; two-server rig, 3-minute guest TTL unless noted).

**Android smoke suite, all on the release candidate**

| Run | Result |
|---|---|
| Federation rig, HTTP parent (proxy checks + direct phase + renewal + key revoked) | 16 / 16 |
| Federation rig, Quick Connect parent (+ parent released/re-dialed, switch, relaunch) | 21 / 21 on the third attempt — the first two never got the parent's fresh Quick Connect tunnel up (six 13 s handshake stalls in a row, mStream#940); a 15 s warm-up after enabling Quick Connect fixed the rig, not the app |
| Heal hand-off: expired token, parent dead → probe → skip walk, parent back → renewal → resume | all pass, walk at ~4 s/track ending cleanly |
| Download cap, keep-queue-offline on | 0 stream-cap rejections (was 23), 26/26 landed, playback undisturbed |
| launch-matrix | 7 / 7 |
| switch-during-outage | 3 / 3 |
| media-resumption | 3 / 3 |
| dead-zone (cellular, two airplane outages, Retry, Wi-Fi return) | 7 / 7, reconnects in place, no rebuilds |

**iOS**

| Run | Result |
|---|---|
| Simulator (iPhone 17 Pro), full direct round | ticket on the switch, own tunnel on the first dial (3.2 s, direct, guest), albums and album-songs 200 over it, 6 queued URLs on the peer's loopback, renewal in place, parent killed → next track + browse OK, parent restarted → second renewal; repeated on the final build |
| Simulator `sim-launch.sh` | boot + default published pass; the resume check timed out because the restored config's Quick Connect server isn't running (the resume line is in the log later) |
| iPhone X | now on iOS 16.7 with Developer Mode off, so nothing can be launched or attached from the Mac. It has a profile build of this stack installed and the rig parent planted first in its server list; the rig pair for it is left running (peer 3111, parent 3112) for a hand test in the morning. Backups of its documents are under `smoke/out/device-backups-20260905/`. |

**Fixed along the way (in #151 / #152)**

- Quick Connect parent: once a peer went direct, nothing reconciled the parent's tunnel away until the next queue edit or switch — a bind now reconciles, and a released tunnel rebuilds whatever still rode it.
- A queue restored at launch sits ready-but-paused; an upcoming-only rebuild now widens to a paused current track as well, so it moves to the peer's loopback instead of staying on a proxy port that may be released.
- The rig: the queued-URL check ignores downloaded copies and polls; a Quick Connect reconcile waits out the first-dial stall ladder; after revoking the key the rig presses NEXT so a buffered track cannot mask the outcome.

**Still open**

- The first guest dial through a freshly enabled Quick Connect endpoint stalls in the handshake for 13 s a time, up to four times tonight (mStream#940). The retry ladder lands it; the simulator never stalls.
- Casting a direct peer to a renderer was not run (a TV in the house at night).

PR #155, opened 2026-09-06
native: refresh the iroh tunnel binaries so a stalled handshake names its path

IrosTheBeggar:
The source landed in #153; this refreshes the committed binaries built from it: Android arm64-v8a and x86_64 (`build-android.sh`), the iOS xcframework for device and simulator (`build-ios.sh`), README checksums updated. The only other change is the xcframework's `Info.plist` reordering its two entries.

Verified on the rebuilt artifacts: `cargo test`; the interop harness ALL PASS (tunnel, kick, guest dial, credential swap with the port kept); the iOS simulator dials the interop pairing server first try (3.2 s, direct) on the new framework; both `.so` files export the 14 ABI v2 symbols and carry the new message. The Galaxy was unplugged before the launch-matrix could run on the new library; the library differs from the shipped one only by this message string.


IrosTheBeggar:
Galaxy S25 on the rebuilt arm64 library (debug `.plus.dev` APK): launch-matrix 7 / 7 (standard default and Quick Connect default), dead-zone 7 / 7 (Wi-Fi to cellular hand-off, two 75 s airplane outages reconnecting in place, the Retry tap kicking in place, Wi-Fi return, no rebuilds). That closes the gap noted in the description.

PR #159, opened 2026-09-07
smoke: the rig taps the peer's real Albums tile; sim-launch documents its resume precondition

IrosTheBeggar:
Two harness tidy-ups from the multi-server Auto DJ round.

- **federation-rig.sh** — the peer's home carries a *Read-only server* note above its grid, which puts Albums top-right (Galaxy `796 842`). The old default (`234 909`) landed on File Explorer and the run passed through a folder row by accident. The new default is the same tap trio `autodj-multi-server.sh` phase 1b used on the Galaxy today (39/39); the emulator override in the header is unchanged.
- **ios/sim-launch.sh** — the resume check reads a line only the Quick Connect transport logs, so the simulator needs a Quick Connect default; the header now says so instead of the script failing 2/3 on a config without one.

Not re-run on the phone (it had locked again); the rig's later steps are unaffected — the back tap then opens the album, which the parent-dead check already accepts.

### DIFF
--- a/smoke/android/federation-rig.sh
+++ b/smoke/android/federation-rig.sh
@@ -33,10 +33,12 @@
 # Everything it creates is torn down: both servers, their scratch dirs, the
 # phone's server list (cfg_restore) and the peer's download folder.
 #
-# Taps are Galaxy S25 defaults. On the arm64 "smoke" AVD (Pixel 7 profile,
-# 1080x2400) the peer's home grid sits lower and a track ROW tap only appends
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
Re-run on the Galaxy S25 with the corrected default: `federation-rig.sh` HTTP **16/16**. The log shows the Albums route this time (`POST /api/v1/db/album-songs → 200` on the album open and again on the parent-dead check), not the File Explorer folder row the old coordinate fell through.

PR #165, opened 2026-09-07
feat(federation): requests waiting — a home banner and a phone notification

IrosTheBeggar:
Builds the Tier 1 plan from the assessment: a banner on the browse page while federation requests are waiting, and a phone notification when one comes in while the app is open or playing. **Stacked on #164** (retarget after it merges). The count comes from the server — **IrosTheBeggar/mStream#957** adds `user.federationInbox` to the boot payload; against an older server nothing shows.

## What the phone gets

- **Home banner.** Above the sections: "N federation requests waiting · Tap to accept or reject". Tapping opens the Federation screen. The screen's own request list writes the count back, so an accept or a decline clears (or rewrites) the banner at once.
- **Notification.** A rise in the count posts "N federation requests are waiting · On <server>. Open to accept or reject."; a drop retires it. Tapping opens that server's Federation screen, also from a cold launch. One notification per server.
- **When it can fire.** On every ping (launch, server switch, resume) and on a 3-minute re-ping that runs only while the app is in front or audio is playing (the Dart isolate is alive for the player then). With the app closed nothing polls — that is the Tier 2 work, deliberately not here.
- **Settings.** "Notify me about requests" in Federation settings (default on). The OS permission is asked for when the inbox or the switch is turned on, or on the first alert while the app is in front — never at launch.

## Model

- `Server.federationInbox` (persisted; null on peers and on servers without the field), read at both ping sites; changes persist and rebuild the home in place.
- `ServerManager.setFederationInbox` — the Federation controller feeds it after reading the request list.
- `lib/singletons/federation_inbox_alerts.dart` — `inboxAlertFor` (rise → notify, drop → clear, same → nothing) and `shouldPollInbox` are pure; the singleton owns the plugin, the timer, the lifecycle hook and the tap-through via a new `rootNavigatorKey`.
- `flutter_local_notifications` 22.3: Android needs `coreLibraryDesugaringEnabled` (its AAR metadata refuses otherwise, even at minSdk 26); iOS hands the notification center's callbacks to the plugin through `FlutterAppDelegate`.
- 8 new l10n keys, all ten locales.

## Tests

`federation_inbox_alerts_test` (alert rule, poll filter), `federated_nav_test` (banner item, still a home list, no NETWORK group from the count alone), `federated_server_test` (JSON round trip). Full suite 591/591; analyzer at the 17-info baseline; `flutter build ios --simulator` green.

## Verified on the Android emulator against the rig (B on the #957 patch)

- a planted received request → launch → banner + "A federation request is waiting" in the shade
- a second one planted while the app was in the background → resume → "2 federation requests are waiting", banner rewritten
- tap on the notification → Federation screen with both requests
- Decline one → count 2→1, notification retired (no live record), banner shows 1
- the settings switch renders; permission was pre-granted with `pm grant`, so the runtime prompt itself was not exercised

## Not covered

- iOS runtime (build only): the permission prompt, foreground presentation and the tap-through on a device.
- Alerts with the app closed (Tier 2).
- A peer's own inbox: the count is deliberately null on peers.


IrosTheBeggar:
**Device smoke, 2026-09-10** — rebased onto master (`ba81a69` + `462f14a`), rig B on mStream master (6.26.0, carries #957), phones on the LAN.

**Galaxy S25 (Android 16), `.plus.dev` debug build:**
- launch with one planted request → banner "1 federation request waiting" + the notification in the shade
- Albums → play, HOME, second request planted on the server → **notified 80 s later with the app in the background and music playing** (the 3-minute re-ping) — "2 federation requests are waiting", banner rewritten
- tap the notification → Federation screen with both requests
- Decline one → count 2→1, notification retired, banner rewritten
- "Notify me about requests" off → third request → resume → banner 3→2 updated, **no** notification; switch back on
- the notification permission had been granted on this phone earlier, so the runtime prompt itself was not exercised here

**Found and fixed on the way:** the rebase (master reindented the settings screen for the SafeArea fix) had dropped the notify switch into the ticket-defaults card; `462f14a` moves it back under the inbox switch.

**iPhone X (iOS 16.7), profile build:** installed with the rig config; launching over USB kills the app on debugger detach (an `lldb_image_notifier` SIGTRAP, not an app crash), so the run needs the icon tapped on the phone — pending.

IrosTheBeggar:
**iPhone X (iOS 16.7), profile build — done:**
- launch with two planted requests → banner, the permission prompt, the first notification (seen by the user)
- a request planted after launch was picked up by the in-app 3-minute poll with the app in front (the saved count matched the server)
- **background case:** music playing, app sent to the home screen, request planted → "4 federation requests are waiting" banner over the home screen **108 s later**; the app saved the count
- tap-through from the shade: pending the user's tap

Both phones now pass the Tier 1 promise: a notification while the app is open or playing in the background.

PR #171, opened 2026-09-12
fix(l10n): grey screen on every non-English device since v0.36.0

### LINKED ISSUE #169: Android app stuck on grey screen – v0.36.0 / Android 16
Hi,

I’m using v0.36.0 on Android 16 (Nothing Phone / Nothing OS 4.1).

Unfortunately, I’m currently unable to use the Android app because it gets stuck on a grey screen.

I’ve already tried clearing the app cache and storage, but the issue persists.

I also have screen recordings showing the problem hosted on my Nextcloud server, in case they’re helpful for troubleshooting:

https://nextcloud.elixs.de/s/HpK29cGCDa3JYZG

Stay safe,
Elias

IrosTheBeggar:
## Summary
Since v0.36.0 the app shows a grey screen on the home and Add Server screens for every device whose language is not English (#169, reported on a German Nothing Phone running Android 16). It is not an Android 16 or Nothing OS problem; v0.35.0 is unaffected.

## Root cause
PR #126 moved the UI to `package:material_ui`, whose `MaterialLocalizations` is a different Dart type from `package:flutter/material`'s. The generated `AppLocalizations.localizationsDelegates` only carries the flutter_localizations delegates (the flutter/material type), so material_ui widgets were left with `DefaultMaterialLocalizations`, which supports only `en`. The app declares de/es/fr/it/ja/pl/pt/ru/zh as supported locales, so a German phone resolves to `de`, Flutter loads no material_ui `MaterialLocalizations` for it, and every `MaterialLocalizations.of(context)` throws: the drawer tooltip on the home screen, `TextField`/`TabBar` on Add Server, the language `DropdownButton` on Welcome. In release that exception renders as the grey `ErrorWidget`. Unsupported device languages fall back to the first supported locale, which is `de`, so they were hit as well. Only English devices escaped.

## Fix
- `lib/l10n/localizations_delegates.dart`: `appLocalizationsDelegates` = the generated list + material_ui's `GlobalMaterialLocalizations.delegate` + cupertino_ui's `GlobalCupertinoLocalizations.delegate` (cupertino_ui becomes a direct dependency for the import).
- Every `MaterialApp` (`main.dart`, `test/`, `integration_test/`) uses that list.
- `test/localizations_delegates_test.dart` resolves all ten supported locales and keeps a negative control proving the generated-only list fails for `de`.

## Verification
- `flutter analyze` clean on the touched files; the full `flutter test` suite passes (622 tests).
- Reproduced with a debug build on Pixel_4_API_31 (app language `de`) and Pixel_6_API_36 (Android 16, per-app locale `de-DE`): red "No MaterialLocalizations found" on exactly the screens in the report. With the fix, home and Add Server render in German on both, no exceptions in logcat.
- Release-mode smoke on the Android 16 emulator (`de-DE`, an unsupported `nl-NL`, and `en-US`): running now, results posted as a comment before merge.

Thanks to @elixs-de for the report and the screen recording. It caught this before the build reached Google Play.

Fixes #169

### DIFF
--- a/integration_test/helpers/test_helpers.dart
+++ b/integration_test/helpers/test_helpers.dart
@@ -35,11 +35,12 @@ import 'package:path_provider/path_provider.dart';
 
 import 'package:mstream_music/main.dart';
 import 'package:mstream_music/l10n/app_localizations.dart';
+import 'package:mstream_music/l10n/localizations_delegates.dart';
 import 'package:mstream_music/singletons/server_list.dart';
 import 'package:mstream_music/singletons/browser_list.dart';
 
 Widget testApp() => MaterialApp(
-      localizationsDelegates: AppLocalizations.localizationsDelegates,
+      localizationsDelegates: appLocalizationsDelegates,
       supportedLocales: AppLocalizations.supportedLocales,
       home: MStreamApp(),
     );
--- a/integration_test/server_edit_jwt_test.dart
+++ b/integration_test/server_edit_jwt_test.dart
@@ -20,6 +20,7 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 
 import 'package:mstream_music/l10n/app_localizations.dart';
+import 'package:mstream_music/l10n/localizations_delegates.dart';
 import 'package:mstream_music/objects/server.dart';
 import 'package:mstream_music/screens/add_server.dart';
 import 'package:mstream_music/singletons/server_list.dart';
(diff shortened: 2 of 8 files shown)


IrosTheBeggar:
Release-mode smoke (signed `flutter build apk --release --flavor full`, fresh install on the Android 16 emulator `Pixel_6_API_36`, per-app locale via `cmd locale set-app-locales`):

- `de-DE`: Welcome (including the language dropdown), Add Server (Quick Connect tab, pairing-code field) and the main screen all render in German. Before the fix all three were grey.
- `nl-NL`: renders in English. A per-app locale keeps the system's `en-US` in the preferred list, so this exercised the English path; an all-Dutch phone resolves to `de`, which is the path above.
- `en-US`: renders as before.
- `adb logcat`: no `E/flutter` and no widget-library exceptions across all runs.

Full `flutter test`: 622 passed. Merging and cutting v0.37.1.

PR #84, opened 2026-06-24
Windows desktop port: build, desktop shell, and iroh tunnel

IrosTheBeggar:
## Summary

Brings mStream Music to **Windows desktop** as a traditional desktop player. The app builds, launches, plays, browses, and now loads the iroh remote-access tunnel — all reusing the existing singletons/streams (no playback or business logic duplicated). A companion `DESKTOP_PORT_PLAN.md` documents the full status and the feasibility/effort breakdown for the remaining desktop features.

## What's in it

**Build enablement**
- Scaffolds the `windows/` runner.
- Routes `just_audio` through **media_kit (libmpv)** on Windows/Linux, where just_audio has no native backend (`just_audio_media_kit` + `media_kit_libs_audio`), initialized in `main()` for those platforms only.
- Gates `AudioService.init` to Android/iOS/macOS (no Windows/Linux platform impl) and constructs `AudioPlayerHandler` directly elsewhere — in-app playback is identical, only OS media-session integration is skipped.
- Drops the portrait-orientation lock on desktop.

**Desktop UI** (`lib/widgets/desktop_shell.dart`)
- A width-breakpoint switch in `MStreamApp.build` selects `DesktopShell` on desktop platforms ≥ 900px, else the unchanged phone shell.
- Persistent left sidebar (server picker · library · tools), browse/album-detail in a nested `Navigator`, a toggleable queue panel, and a full-width Now Playing bar (art · transport · seek · volume · queue).

**iroh remote tunnel on desktop**
- `rust/iroh_tunnel` builds for `x86_64-pc-windows-msvc` **unchanged** — its Android-only deps (`jni`, `ndk-context`) and `android_init` are already `cfg`-gated.
- The prebuilt `iroh_tunnel.dll` is committed under `windows/iroh/` and bundled next to the exe via an `OPTIONAL` CMake install rule, mirroring the Android `jniLibs` prebuilt-`.so` convention.
- `iroh_tunnel.dart` loads the platform-appropriate library name and allows desktop in `isSupported`. `add_server._scanQr` now gates QR scanning on the camera platform (Android/iOS) rather than `isSupported`, so desktop falls back to the existing paste-the-code path.

## Verification (Windows)

- `flutter analyze` — clean
- `flutter build windows` (debug + release) — succeeds; ~51 MB release bundle
- Boots into the desktop shell with **no layout overflow**
[...]

PR #173, opened 2026-09-12
feat(storage): per-server mirror root — play from a folder another tool keeps in sync

IrosTheBeggar:
## Summary

A1 of `BACKUP_SYNC_IMPLEMENTATION.md` (Tier A of the backup & sync plan). Adds an optional per-server **mirror root**: a folder that already holds a copy of the library in `<vpath>/<rel>` shape, kept current by something other than the app (Syncthing, rclone, robocopy, a NAS mirror). Files found there count as local copies — playback uses them, the downloaded badge shows, and downloading a track that is already mirrored is a no-op. The app never writes there.

- **`Server.mirrorRoot`** — persisted; absent/non-string → null (old `servers.json` files load unchanged).
- **`lib/util/local_copy.dart`** — `localCopyCandidates()` is now the one place that knows both on-disk shapes: the app's download tree first (`<dir>/media/<localname><data>`, formula unchanged), then `mirrorRoot + data`, normalised so a Windows root and the server's `/` separators make one native path. `firstExistingSync()` picks the first that exists. The four resolution sites go through it: the play-time MediaItem build (`queue_actions`), the batch queue fill (`api.dart`), the badge recheck (`DisplayItem.recheckDownloadedIn` — now also checked when the download location is unavailable), and `downloadOneFile`'s already-on-disk check.
- **Add / Edit Server** — the one visual change: a "Local copy folder (optional)" row with *Choose folder* and *Clear*, under the storage section. Offered on desktop (native directory picker via `file_selector`) and on the full Android flavor (all-files access + the existing in-app folder browser — the same gate and picker as Permanent). Hidden on iOS and the Play build. Saving an edit re-checks the browser's badges, as a storage change already did.
- Strings in all ten ARBs; generated l10n committed.

## Test plan

- [x] `flutter analyze` — no new issues (the 21 reported are all pre-existing and untouched).
- [x] `flutter test` — 636 pass, including the new `test/util/local_copy_test.dart` (candidate order, unavailable download dir, unset/empty root, Windows normalisation, first-existing) and `test/objects/server_mirror_root_test.dart` (round trip, absent, non-string, cleared).
- [ ] Device smoke still to do: S25 `--flavor full` with a robocopy'd folder under `/sdcard/Music` → badge + local playback; desktop picker once the desktop shell (#84) lands.

Note: the all-files-access prompt reuses the existing "store downloads permanently" wording.


IrosTheBeggar:
**Emulator smoke (Pixel_4_API_31 · Android 12 · full flavor · feat/library-sync @ 9eb55d0 · mStream v6.27.0 on the host):**

- Seeded `mirrorRoot` = `/storage/emulated/0/Music/mstream-mirror` with one track planted there (`music/Vosto/Night Drive/02 - Neon.mp3`, not in the app tree): the row shows the downloaded badge; tapping it plays with `extras.localPath` resolved to the mirror file (queue.json), neighbours have none.
- Folder "Download" on that album transferred only Highway + Static into the app tree — the mirrored Neon was skipped (no-op), snackbar counted 3 as before.
- Edit Server: the "Local copy folder (optional)" row renders (label, help, Choose folder, path, Clear); Choose folder opens the in-app browser with all-files access; picking a folder updates the row; Save persists it; Clear + Save persists `null`; after clearing, the Neon badge disappears on the next browse while the two real downloads keep theirs.
- No Flutter exceptions in logcat.

Not covered: desktop picker (needs the desktop shell), physical device.

PR #177, opened 2026-09-12
feat(mirror): Library copy — MirrorManager, transport adapters, the screen

IrosTheBeggar:
## Summary

Second half of A3 (`BACKUP_SYNC_IMPLEMENTATION.md` §4.3–4.4): the app around the mirror engine from #176, and the one visual change — a **Library copy** screen.

- **`MirrorManager`** (`lib/singletons/mirror_manager.dart`): runs the engine for every server with a rule — 2 min after boot, every 6 h while the app is open, and on demand — one run per server at a time, with a status stream for the UI. Cancels stale mirror-group transfers left by a previous process at init. Refreshes the browser's badges after each run.
- **Transport adapters**: `ServerManifestClient` posts `sync/manifest` over the server's API transport (HTTP or the iroh loopback) with the JWT header and a quoted `If-None-Match` for the 304 fast path — its own client, so the browser's loading bar never notices. The engine's `Downloader` maps to background_downloader tasks in a dedicated `mirror` group (resumable, 5 retries, Wi-Fi-gated on request, priority below manual downloads) that land in the runner's temp tree; `DownloadManager._onUpdate` hands that group's updates to the manager instead of looking for a tracker.
- **Server model**: `syncAvailable` from the `/api/` features block (flags, never probes — the screen tells you when a server is too old), `mirrorRetentionDays` (30) and `mirrorWifiOnly` (true), all persisted.
- **Library copy screen** (Manage Servers → ⋮ → *Library copy*): one switch per library ("Keep a full copy on this device" — original quality only, per the decision), a status card (last sync · files · size, progress while running, last error, a tappable count of failed files with their errors), *Sync now* / *Cancel*, trash retention (7 / 30 / 90 days / forever) and, on mobile, *Wi-Fi only*; desktop gets a note that syncing happens while the app is open. Turning a library off trashes what no rule wants any more — recoverable for the retention period.
- Small extras: three status queries on the index (`localBytes`, `localCount(state:)`, `localFailed`); the byte formatter moved out of `add_server.dart` into `lib/util/format_bytes.dart`; 18 strings in all ten ARBs.

Deliberately not in this PR: the free-space preflight hook (the engine supports it; the app passes none yet), moving the play-time resolver onto the index (the mirror writes into the same `media/<server>` tree the resolver already checks), and the entity rules (A6).

## Test plan

- [x] `flutter analyze` — no new issues; package `dart test` 53/53; `flutter test` 647/647 incl. `test/singletons/mirror_manager_test.dart` (manifest client: token + quoted tag + paging body, 304 → null, non-200 throws; rule toggling publishes status and a server without the flag never runs; model defaults / round trip / junk fallback).
- [ ] Emulator smoke (in progress): enable the copy of the fixture library against the local 6.27.0 server → files land in `media/<server>`, status line, badges.


IrosTheBeggar:
**Emulator smoke (Pixel_4_API_31 · Android 12 · full flavor · this branch @ 4bda0c5 · mStream v6.27.0 on the host, 9-track fixture library):**

- Boot: `syncAvailable: true` persisted from the `/api/` features block; two leftover downloads imported as `manual`.
- Manage Servers → ⋮ → **Bibliothekskopie** renders (section, help, `music` switch, status, Sync now, retention 30 d, Wi-Fi only).
- Switching `music` on started a run immediately: `[mirror] smoke: ok — 7 new, 0 replaced, 0 renamed, 0 trashed, 0 failed, 2 unchanged`. All 9 files under `media/smoke/…`; the two pre-existing downloads were **adopted** (origin kept `manual`, hash recorded, mtime stamped), not re-downloaded. 9 required-by edges, run row `rule / 7 downloaded / 2 unchanged`, status card "Zuletzt synchronisiert … · 7 Dateien · 59.7 KB". No temp or trash leftovers.
- Server side: renamed one fixture and deleted another, rescanned (`admin/db/scan/all`), manifest revision moved. **Sync now** → `0 new, 0 replaced, 1 renamed, 1 trashed, 0 failed, 7 unchanged`: the renamed file was moved locally with zero transfers; the deleted one sits in `.mstream-trash/smoke/2026-09-12/music/Icarus/Later Works/02 - Descent.mp3`; index at 8 rows.
- No Flutter exceptions in logcat.

Not covered: cancel mid-run, Wi-Fi gating on a metered network, desktop.

PR #180, opened 2026-09-13
feat(mirror): album and artist "Keep offline" rules (A6a)

IrosTheBeggar:
## Summary

A6a of `BACKUP_SYNC_IMPLEMENTATION.md`: **"Keep offline" for an album or an artist** — entity rules for the library copy, next to the whole-library switch from A3.

- **Engine**: the runner expands rules by kind through one pure function, `expandRule`. `library` and `folder` match by path as before; `album` is every track tagged with that album name (the identity the album screens use); `artist` is every track by them **plus the albums credited to them as album artist** (`remote_albums`), so a compilation that merely contains one of their tracks is not pulled in whole. `RuleKind` names the kinds. Removing a rule works exactly as for the library switch: the next run re-plans and trashes only what no remaining rule pins.
- **App**: `MirrorManager.keepsRule / setRule / canKeep` (the library switch is now the `library` kind of the same); a rule toggle syncs right away, like before.
- **The one visual change**: an offline-pin toggle in the album detail toolbar (filled while the album is pinned), and the same switch on an artist row's long-press sheet. Both appear only on servers with library sync (`canKeep`). Three strings in all ten locales.

Not in this PR: playlist / rated rules and a place to see every rule at once (A6b), the transcoded tier (A7), per-rule Wi-Fi (the server-level Wi-Fi setting still gates the whole run).

## Test plan

- [x] Package `dart test` 67/67: `expandRule` per kind (prefixes, album by tag, artist by tag or credited album, unknown kind pins nothing); a run with an album rule downloads only that album, an artist rule adds their track on another album, dropping the album rule keeps what the artist rule still credits and trashes the guest track once the credit changes.
- [x] `flutter analyze` at the 21-issue baseline; `flutter test` incl. the new MirrorManager rule test (`canKeep`, `setRule` on/off, `keepsRule`).
- [x] Emulator smoke on Pixel_4_API_31 (results in the comment below): library rule off → trashed; album pin on → its tracks downloaded, pin filled; artist switch on → the artist's other tracks; album pin off → nothing trashed while the artist rule covers it; artist off → trashed.


IrosTheBeggar:
## A6a emulator smoke — PASSED

Pixel_4_API_31 (German locale), full-flavor debug APK at 06988fa, local mStream 6.27.0 on port 3580, starting from the whole-library rule (9 files mirrored, one of them a manually downloaded copy the mirror had adopted).

| # | Step | Result |
|---|---|---|
| 1 | Library copy → "music" switch off | ✅ run: `7 trashed` (the manual-origin copy stays, as designed) |
| 2 | Alben → "Be Somebody" → toolbar pin (outlined) | ✅ `[mirror] smoke: ok — 2 new, … 1 unchanged`; pin filled |
| 3 | Künstler → long-press "Icarus" → "Offline behalten" on | ✅ sheet with the help text; run: `2 new, … 3 unchanged` (the other album's two tracks) |
| 4 | Album pin off while the artist rule stays | ✅ run: `0 trashed, 5 unchanged` — the artist rule still pins the album's tracks; pin outlined again |
| 5 | Artist switch off | ✅ run: `4 trashed` (the four mirror-owned files; the adopted manual copy is left alone) |

Not covered: an artist credited as album artist on an album whose tracks carry other artists (unit-tested), desktop, physical device.

PR #183, opened 2026-09-13
fix(mirror): wire the free-space preflight into the app's runs

IrosTheBeggar:
## Summary

A3 follow-up found while preparing the Galaxy test: the engine's free-space preflight (`MirrorRunner.freeSpace`, refuses a run when `free < bytesNeeded + 64 MiB`) was never given a probe by the app, so a whole-library rule on a phone without the room would have filled the disk until downloads failed. `MirrorManager` now passes the same StatFs call the storage migration uses (`MigrationManager.freeBytes`). The refusal surfaces as the run's error on the Library copy screen ("Last sync failed: not enough free space: need … have …").

## Test plan

- [x] `flutter analyze` at the 21-issue baseline; `flutter test` (the preflight itself is covered by the package's runner test "preflight refuses a run that will not fit").
- [ ] Galaxy: a whole-library rule for the real library is refused with the message instead of filling the phone.


IrosTheBeggar:
## Galaxy S25 smoke — feat/library-sync at b62a3e2, real server over iroh — PASSED

SM-S931U, Android 16, full-flavor release build, server `iroh://304d4676` (mStream 6.27.0, signed in). The whole-library switch stayed off: the phone cannot hold the full library.

| # | Step | Result |
|---|---|---|
| 1 | Boot ping over the tunnel | ✅ drawer shows "Server v6.27.0"; the boot+2 min run refreshed the index (`ok — 0 new … 0 unchanged`, no rules) |
| 2 | Library copy | ✅ one vpath ("library", off), status, offline switch, rated dropdown, empty rules list |
| 3 | Albums → "'Round About Midnight" → toolbar pin | ✅ `[mirror] iroh-304d4676: ok — 6 new` in ~25 s over the tunnel; pin filled; Library copy: "6 files · 67.1 MB", the album in "Also kept offline" |
| 4 | "Browse the offline copy" on | ✅ chip; the Albums grid comes from the index for the whole library; search "midnight" → grouped albums / tracks with the pill and no API call |
| 5 | Rate "'Round Midnight" ★★★★ offline | ✅ `[outbox] iroh-304d4676: queued rate` |
| 6 | Play the track while offline | ✅ plays from the mirror copy |
| 7 | Switch off | ✅ `[outbox] iroh-304d4676: replayed 1 change(s)` six seconds later; Rated refetched online (200) |

Not tested: the free-space refusal (#183) — it needs a library larger than the free space; auto-offline (cutting Wi-Fi would also cut the adb link). Left on the phone: the album pin (67 MB) and the 4-star rating on the server.

PR #197, opened 2026-09-19
feat(federation): fold the parent's `direct` hint into the denial state (Phase 8d)

IrosTheBeggar:
## Summary

The client half of FEDERATION_PLAN Phase 8d, against mStream #1003 (merged): the parent's peer listing now carries a per-peer `direct` hint — `true` when the parent holds a guest token (the peer mints), `false` when the peer refused the parent's last mint, `null` when never asked — and the app folds it into its denial state during the peer reconcile instead of learning every refusal one access call at a time.

**The rule** (`ServerManager.applyDirectHint`, pure, unit-tested; applied by `_reconcilePeers` on new and existing peers):

- `true` lifts any denial: the parent holds a token, so the peer mints.
- `false` files a denial once — on the hint's first appearance or a flip from `true` — and only when the client holds none. A denial of its own keeps its more precise time, and a *steady* `false` leaves an aged-out denial alone, so the hourly re-ask (8b) still reaches the parent and refreshes **its** memory. Without that, a peer upgraded to a minting build would be rediscovered by nobody: clients trusting the hint would never ask, and the parent only re-learns when asked.
- `null` changes nothing: never asked, or a parent too old to say.

A denial lifted or learned re-runs `ensureTunnels('direct-hint')`, so a peer the parent now knows mints gets its tunnel at once. `Server.federationDirectHint` is runtime-only; nothing persisted moves. Older servers omit the key and the app behaves exactly as before.

**What it buys.** No access call and no bridge dial for a declined peer at launch or on a switch, on every device; the parent's knowledge is fresher than any one client's; and the first dial for a peer the parent knows mints is certain.

## Verification

- `flutter analyze` clean in the touched files; the full suite passes (739 tests).
- The rule's eight cases are pinned in `test/objects/federated_server_test.dart`; the runtime-only field is checked not to persist.
- Not device-run: the rig's peer always mints, so only the `true` path can be exercised there, and it is a no-op for a client with no denial. The `false` path is the unit test's.


IrosTheBeggar:
## Emulator smoke — arm64 AVD, the merged server (mStream #1003 in `~/code/mStream`'s worktree)

The federation rig (HTTP mode) with two new assertions for the server change: the peers listing says `direct: null` before anyone asks and `true` once the parent has cached the token.

| Run | Result | Note |
|---|---|---|
| 1 | 22 pass, 1 fail | The lapse leg's `ensure_playing` sent a PLAY key while the reload was still in flight; audio_service folds every external play/pause key into the media toggle, so it paused the resume it was waiting for. The app's renewal itself was fine (`guest credential refreshed in place (401)`). → b128eda: the leg polls (`wait_playing`), presses nothing. |
| 2 | 22 pass, 1 fail | Lapse leg green. The download-folder check failed because an older rig session had left a `rig-b` at `10.0.2.2` and its peer record on this emulator, so the reconcile adopted the old record instead of creating one. Stale device state; cleaned. |
| 3 | **23 pass, 0 fail, 0 skip** | Both hint assertions, the direct path, the renewal, the lapse renewed from the playback path, the revocation leg. |

The `false` hint path cannot be produced on the rig (its peer always mints); it is pinned by the unit test. Full unit suite: 739 passed.

PR #200, opened 2026-09-19
feat(iroh): the startup diagnostic names the tunnel crate's version

IrosTheBeggar:
`mstream_iroh_version()` arrived with the tunnel crate's 0.2.0 (additive; the ABI stays 2). The binding looks it up as an optional symbol — a v0.1.x binary simply lacks it and reads null — and `IrohTunnel.version` exposes it. A line right after the app's startup line now reads

```
[iroh] native tunnel supported: true (ABI v2, crate 0.2.0)
```

("crate pre-0.2.0" for an older binary; the reason when unsupported), so a diagnostics export names the exact binary that shipped, not only what `tool/iroh-tunnel.version` expected the checkout to carry. The desktop branch carries a line of the same shape without the crate; its rebase takes this one.

Verification: the Android smoke suite on the Galaxy S25 with this build, plus the startup line read from its log — results to follow on this PR.


IrosTheBeggar:
Verified on the Galaxy S25 with this build (`.plus.dev`, v0.2.0 binaries). The startup line reads

```
[app] mStream v0.38.0 started
[iroh] native tunnel supported: true (ABI v2, crate 0.2.0)
```

The whole Android chain, 12 legs, 0 failures:

| leg | result |
|---|---|
| session-actions | 2 pass |
| launch-matrix | 6 pass |
| federation-rig (iroh) | 26 pass, 1 skip |
| federation-rig (http) | 21 pass, 1 skip |
| qc-first-dial (3 iterations) | 1 pass |
| switch-during-outage | 2 pass, 1 skip |
| media-resumption | 3 pass |
| dead-zone | 7 pass |
| autodj-multi-server (default cap) | 37 pass, 1 skip |
| autodj-multi-server (`SMOKE_DJ_MAX_STREAMS=3`) | 37 pass, 1 skip |
| play-history (`SMOKE_HOST` = the Mac's LAN address) | 21 pass |
| playback-soak (10 min) | 2 pass |

The skips are the scripts' own preconditions (the hint legs on a server checkout that predates the direct-hint PR; a queue not on the Quick Connect server, so the tunnel was released on the switch instead of coming back; a DJ pick from the peer whose queued track keeps its tunnel). The cap-3 run made parent A answer 429 eleven times, but the leg skips tracks every 1.6 s, faster than the player surfaces a failed load, so the app's 429 back-off still never engaged on a device; that needs a leg that holds the cap with parked streams while one direct track plays.

PR #202, opened 2026-09-21
Android: a Now Playing home-screen widget

IrosTheBeggar:
## Summary

PR 1 of the home-screen widget plan: the Android side. Cover, title, artist and previous / play-pause / next for the track the audio handler is on, plus an empty state that opens the app.

- **`packages/now_playing_widget`** — an in-repo path plugin (same reason as `wifi_lock_shim`: it must register on the headless engine, or the widget goes stale after a Bluetooth-key / media-resumption boot). RemoteViews with a 4x1 row layout and a taller one from two cells up, the wallpaper palette on Android 12+ and dark mode below it, strings in all eleven app languages.
- **Dart publishes, never listens** — `NowPlayingWidgetPublisher` sends a debounced, deduped snapshot (item, playing, locale tag) from `main()`, so headless boots keep it current. The native side persists it and re-renders after process death without Dart, trusting "playing" only while audio_service's service is actually running.
- **Buttons drive the media session natively** — a receiver in the package uses audio_service's session transport controls. Binding the MediaBrowserService starts it, so a tap on a dead app cold-boots the handler, and the action is held until the handler is up (the plugin's first publish, or a skip action on the session, which audio_service's baseline never carries). Media-key broadcasts were tried first and rejected: audio_service's explicit-play key (`KEYCODE_MUTE`) is no longer a media-session key on current Android, and a genuine PLAY key folds into the handler's toggle, which is wrong whenever the widget's state is stale.
- **Art through `ArtContentProvider`** — the existing download / cache / host-check / self-signed path, capped at 512 px, filed under a token-blind key (the URL with its session token is never persisted). The provider now also accepts loopback hosts (a Quick Connect server lives at `http://127.0.0.1:<port>` while `servers.json` holds its iroh URL, so its covers were refused) and the federation `/art/` proxy path. Android Auto's covers gain the same two cases.
- **Smoke** — `smoke/android/home-widget.sh` (opt-in in the suite with `SMOKE_WIDGET=1`, since it leaves a widget on the home screen) and `smoke/recipes/home-widget.md` for what needs a phone.

## Verified

- `flutter test`: 748 pass (8 new, publisher). Plugin JUnit: 6 pass (`:now_playing_widget:testDebugUnitTest`). `flutter analyze`: no issues in the new files (the 20 reported are pre-existing elsewhere).
- Emulator (API 36, `.plus.dev` build) smoke 14/14: placement via the debug hook + launcher dialog, launcher shows the published title, Pause / Play / Next (or Previous at the queue's end) reach the handler and the icon follows, a cover served from the Mac is fetched through the provider and rendered, a Play tap after `kill -9` cold-boots the service headless and plays once the handler is up.
- By hand on the emulator: dark mode, and the app's language override (Japanese button descriptions, English again after the revert).

## Not verified

- The taller layout on a launcher resize (the render path is shared; only the layout choice by `OPTION_APPWIDGET_MIN_HEIGHT` is untested visually).
- A real phone (Galaxy S25): the launcher's placement dialog wording, Quick Connect and federated-peer covers on device, the paused state after a reboot — all in the recipe.
- iOS is PR 2 (WidgetKit extension, App Group, intents, URL scheme); the Settings "Add widget" button and lock-screen family are PR 3.


IrosTheBeggar:
Galaxy S25 (SM-S931U, Android 16 / One UI 8.5) round with the `.plus.dev` debug build:

- `home-widget.sh` 13/13: placement through One UI's dialog (its title "Add to Home screen?" matched before the "Add" button on the first try — fixed in e09a6240 on this branch), the launcher shows the published title, Pause / Play / Next reach the handler and the icon follows, the real track's cover renders, a Play tap after `kill -9` cold-boots the service headless and plays.
- `session-actions.sh` 2/2, `media-resumption.sh` 3/3, `launch-matrix.sh` 6/6 (standard and Quick Connect defaults).

Screenshot: the widget on the home page next to the Audible widget, One UI's dark palette, cover from the user's server.

Still not exercised on device: a Quick Connect or federated-peer cover (the queue was on the standard server), the taller layout, the paused state after a reboot.

IrosTheBeggar:
Sizes landed (commit on this branch): one resizable widget with five layouts — 2x1 mini, 2x2 tile (cover fills the card), 4x1 row, 4x2 card, and 4x3+ large with the album line, a natively advancing progress bar with times, and shuffle / repeat. The two live-meter designs were dropped on cost.

- Layout choice is by the frame's aspect ratio plus one width cut-off, because launchers disagree on cell size (Pixel-style ≈ 170 dp for two cells, One UI on the S25 ≈ 140) and absolute thresholds mis-sized the 2x2 on the emulator. Android 12+ gets a sized RemoteViews map; older gets the portrait minimum.
- Progress: Dart sends a position fix and only re-sends on a jump; a 5 s native ticker draws the bar while playing, screen on, and a large widget placed.
- Tests: 755 Dart (7 new), 16 JUnit (new LayoutTest with the three launchers' real sizes).
- Emulator (API 36): smoke 24/24 with every layout forced, progress advancing, mode round trips, cold boot; plus a launcher resize walk through 4x1 → 4x2 → 4x3 → 2x3 → 2x2 → 2x1 confirming the reported sizes and the layout each gets, with screenshots.
- Galaxy S25 / One UI 8.5: 21/23. All layouts, progress and modes pass on the phone. The cold-boot resume stalled: the tap booted the service, the handler logged the play, but the restored queue was on the Quick Connect server and the tunnel never reported up after the `kill -9` — a tunnel cold-dial observation, not a widget one (yesterday's run with a standard-server queue passed).
- Smoke script: pages through the home screen to find the widget, and starts playback with the widget's Play (a PLAY key on this phone went to Audible).

IrosTheBeggar:
The cold-boot resume failure from the 2026-09-21 Galaxy round is not this PR's: the headless process was frozen by One UI's cached-app freezer ten seconds after starting while the play waited on a Quick Connect restore. Fixed in #203 (the play holds the service in the foreground for the wait); with both branches on the phone the widget smoke is 23/23 and the resume 6/6.

IrosTheBeggar:
Two follow-ups from the phone review (commit `2b5814ac`):

- **One picker entry per size.** The picker showed a single 4x1 and hid the rest behind a resize. Now 2×1, 2×2, 4×1, 4×2 and 4×3 are each their own provider with a preview of their own layout (the 4×1 keeps the original class, so widgets placed before the split stay valid). Verified in both pickers: the Pixel launcher lists all five with its size captions; One UI lists "mStream Plus — 5" with name, size and description per entry.
- **The 4x2 looked off.** On the S25 two rows are 168 dp, which already holds the large layout, and the 3x2 that happened to get it looked right while the wider 4x2 got the card (a fixed 88 dp cover and cramped buttons floating in the space). A wide frame now takes the large layout from 165 dp up, whatever its width; the card serves frames between one row and that (small-cell launchers, rotated frames), and it now fills its height too (cover grows with the frame, text at the top, bigger buttons at the bottom). The 4x2 on the phone now shows cover, three lines, progress and the five buttons.

Smoke, on the test build with #203 merged in: Galaxy `home-widget.sh` 23/23 and `media-resumption.sh` 6/6 (Quick Connect queue); emulator 24/24 and 6/6. JUnit 16/16 with the layout table updated to the S25's measured cells (312 × 72 dp, two rows 168).

PR #203, opened 2026-09-21
A cold-boot play holds the service in the foreground while the restore runs

IrosTheBeggar:
## Summary

A headless boot (widget tap, Bluetooth key, media-resumption chip, Android Auto) starts the process with no activity, and audio_service only enters the foreground when `playing` flips to true. Until then the process is cached, and One UI freezes a cached process ten seconds after it starts.

Seen on the Galaxy S25 (One UI 8.5) while smoking the widget PR: `am_freeze` at +10 s with the play still waiting for the queue restore, the Quick Connect tunnel already up in 0.3 s natively, then `sync unfroze` nine minutes later when the widget was tapped and every pending timer fired at once. A restore that dials a tunnel first can take longer than ten seconds, and the handler's wait allows a minute.

**Fix.** In `play()`, when the queue is empty and the restore has not settled, publish the current state with `playing: true` before waiting. That is the transition that makes audio_service call `startForeground` and take its wake lock, and it lands inside the window in which Android still allows a background app to start a foreground service after a user-driven event. The processing state is deliberately left alone (audio_service stops the service on a transition into idle). After the wait, the backend's state is broadcast again, so a restore that never settles leaves a paused notification, not a stuck playing one. Android only.

**Smoke.** `media-resumption.sh` now reports which server the saved queue was on, checks for the hold (or notes the restore beat the play), and reads `logcat -b events` to assert the new process was never frozen before it played.

## Verified

- 744 Dart tests pass (4 new for the pure hold-state helper); analyzer unchanged.
- Device round follows in a comment (Galaxy S25 with a Quick Connect queue, plus the emulator).


IrosTheBeggar:
Device round, on a test build carrying this fix plus the widget branch (#202):

**Galaxy S25 (One UI 8.5), saved queue on the Quick Connect server — the case that stalled yesterday:**
- `media-resumption.sh` 6/6: booted headless on the PLAY key, play reached the handler, **held in the foreground for the restore**, queue restored and playing in 4 s, and the freezer's event log shows the new process was never frozen.
- `home-widget.sh` 23/23, including the widget's cold boot that failed before the fix.
- `launch-matrix.sh` 7/7, `session-actions.sh` 2/2.

**Emulator (API 36, standard-server queue):** `media-resumption.sh` 6/6 (the hold engaged here too, the play still beat the restore), `home-widget.sh` 24/24.

For reference, yesterday's stalled run: `am_proc_start` 11:14:01 → `ActivityManager: freezing` 11:14:11 → `sync unfroze` 11:23:22 on a widget tap, with the Dart log silent in between and every timer firing at once afterwards.

